PR #116, opened 2023-10-14
Render emails to logger if no Sendgrid API key

timglorioso:


src/jobs/send-mail.ts
```diff
@@ -12,42 +12,43 @@ const viewsDir = path.join(__dirname, "../emails");
...
-            relativeTo: path.resolve(viewsDir),
-          },
-        },
-        transport: nodemailer.createTransport(
-          sendgrid({
-            apiKey: process.env.SENDGRID_API_KEY,
-          })
-        ),
+      },
+      transport: !!process.env.SENDGRID_API_KEY
```
timglorioso, line 29: Instead of requiring the API key up front, check it when setting the transport. Note that with this control flow `Email.send` can still be called in a development environment, although currently it's only called in production.

simonv3: So should we remove the production check before `email.send` below and just have it be a json output instead of logging the rest of the e-mail?

timglorioso: Can you clarify with a proposed code snippet? I'm not sure of your proposal. My goal with this PR was to see emails come through the log so I could confirm new accounts in a dev env, without having the Sendgrid API key. It's possible there's a better solution that doesn't involve code changes, but feels valuable for dev env to work without external services

simonv3: For sure, I think I just oversaw that scenario because I had a string in the .env file. 

I'll comment where I think the change could happen, but it's also fine without it.

simonv3 (approved):

src/jobs/send-mail.ts
timglorioso: I'll have to test that it does log with `.send` (unless you've tested it)

simonv3: (I haven't!)

simonv3:
merged cause either way it's fine!

PR #154, opened 2023-10-23
fix broken docs endpoint

ismailghedamsi:


src/index.ts
```diff
@@ -135,7 +135,7 @@ initialize({
 app.use(
   "/docs",
   swaggerUi.serve,
-  swaggerUi.setup(undefined, {
+  swaggerUi.setup(apiDoc, {
```
simonv3, line 138: Hmm, so this loads the static swagger.json into `setup` whereas when it's undefined it reads dynamically from the `req` object [as described in the swagger-express module](https://www.npmjs.com/package/swagger-ui-express#modify-swagger-file-on-the-fly-before-load)

Issue #101, opened 2023-09-29
commitizen + husky on commit


simonv3:
Cause I keep on forgetting to clean up imports

Chickensoupwithrice:
Hey I hope you don't mind me taking this on, since we spoke about it in Discord a while ago. 

I'm partial to using [Pre-commit](https://pre-commit.com) so I was planning on using that to configure these hooks, if that's okay.

jarrodlombardo:
@Chickensoupwithrice Should this be closed? I think you did it with that merge back in October.

simonv3:
@jarrodlombardo I don't think this actually runs a prettier check on commit right now.

Cwpute:
Do we still want Husky and Commitizen ? Should we try to find other tools that do the same thing but better ? Or should we close this issue and move on ?

PR #162, opened 2023-10-25
feat: add husky and commitizen through pre-commit

Chickensoupwithrice:
I'm partial to using [pre-commit](https://pre-commit.com) as mentioned in https://github.com/funmusicplace/mirlo/issues/101 

Currently, I've disabled the husky pre-commit that runs `yarn test` as the tests were failing on my machine (which would not allow me to commit). Let me know if I've done something wrong. I only ran `yarn` to grab all the dependencies. 

In order to to ensure that husky runs `yarn test` before commit, we can un-comment the last line in `.husky/pre-commit` 

I have not used Commitizen before but [this](https://github.com/commitizen/cz-cli/issues/801) issue is where I've copied the pre-commit config from. It appears to be working?

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+# yarn test
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,13 @@
+repos:
+- repo: local
+  hooks:
+    - id: husky-run-pre-commit
+      name: husky
+      language: system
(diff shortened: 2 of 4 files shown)


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
```
simonv3, line 2: does this file exist?

simonv3 (approved):
Husky has changed a bit since I last used it. So when you run `yarn install` this will install the husky hook? 

The tests will fail cause they require a separate docker environment (happy to discuss a better way of doing that?) as detailed in the readme.

PR #168, opened 2023-10-27
Add missing broken translation

ismailghedamsi:


client/src/components/TrackGroup/Buy.tsx
```diff
@@ -33,11 +33,11 @@ const BuyTrackGroup: React.FC<{ trackGroup: TrackGroup }> = ({
     <>
       {trackGroup.minPrice && (
         <>
-          Price: <Money amount={trackGroup.minPrice / 100} />, or
+          {} <Money amount={trackGroup.minPrice / 100} />, or
```
simonv3, line 36: Should this have a translation value?

src/index.ts
```diff
@@ -135,7 +135,7 @@ initialize({
 app.use(
   "/docs",
   swaggerUi.serve,
-  swaggerUi.setup(undefined, {
+  swaggerUi.setup(apiDoc, {
```
simonv3, line 138: Same issue as in the previous PR, I think this generates a static swagger doc.

client/src/components/Home.tsx
```diff
@@ -173,67 +174,69 @@ function Home() {
...
+                  </em>
+                )}
+                <PostContent content={p.content} />
+              </Box>
+            ))}
+          </>
+        )}
+      </div>
+    );
+  }
```
simonv3, line 240: Is there a reason for these indentations?

Might be useful to have prettier installed in your code editor to convert on save, that way it won't create these large diffs. Though it would be even better if it was installed in the repo. I'll fix that.

simonv3:
AMAZING! thank you!!

Looks like the code I _just_ merged caused some conflicts - do you mind trying to fix those? If not I can take a stab at it.

ismailghedamsi:
Yes, i can resolve the conflicts

ismailghedamsi:
I did yarn add id3-parser but i still get the error "Cannot find module 'id3-parser' or its corresponding type declarations"

simonv3:
@ismailghedamsi ah you'll want to run `yarn install` in the `client` directory.

ismailghedamsi:
It works now

simonv3:
K, once that `src/index.ts` `apiDoc` thing is fixed it looks good to merge to me :)

ismailghedamsi:
I don't know how to fix it

src/index.ts
```diff
@@ -136,7 +136,7 @@ initialize({
...
```
simonv3, line 139:
```suggestion
  swaggerUi.setup(undefined, {
```

If this the swagger doc thing a problem for you the ticket should stay open and we can continue looking into it, but with this fix it doesn't work as it should anymore, so I think it hides the problem (and would break the production app).

simonv3:
Closing cause #178 fixes that thing

PR #178, opened 2023-10-31
Add missing broken translation

simonv3:
@ismailghedamsi this is your PR without the apiDoc fix attempt. Gonna merge it so that we can split the issues!


ismailghedamsi:
Perfect

PR #206, opened 2023-11-08
Updated the Readme with a solution to apple chip errors w/ docker

bentooth:
These are the steps used to fix the "error relocating" errors seen when attempting to run docker with an apple chip. big up to timglorioso for the lead


simonv3:
Thank you!!

PR #251, opened 2023-11-17
Update Header.tsx

LouisLouiseKay:
Header didn't display as transparent in dark mode on mobile.


client/src/components/Header/Header.tsx
```diff
@@ -100,6 +100,10 @@ const Header = () => {
 
           @media (prefers-color-scheme: dark) {
             background-color: #232428;
+
+            @media screen and (max-width: ${bp.medium}px) {
+          background-color: transparent;
```
simonv3, line 105: @LouisLouiseKay do you have prettier installed?

PR #255, opened 2023-11-18
Update TrackGroup.tsx

LouisLouiseKay:
Removed borders because they appear even when about and credits are empty


simonv3:
> Removed borders because they appear even when about and credits are empty

This probably is a sign the border needs to be moved to a component deeper in the tree?

PR #259, opened 2023-11-18
Fixed player buttons being unlegible in some contexts.

LouisLouiseKay:
Players buttons had wrong behaviors on hover and depending on user stylings, they're now consistent and mainly adjust to white and dark theme.


LouisLouiseKay:
Added a bunch of other fixes to : 

- Buy button
- Track Row Paddings and Buttons
- Overall Paddings

PR #269, opened 2023-11-19
Update WishlistCollection.tsx

LouisLouiseKay:
Added paddings to wishlist page

### DIFF
--- a/client/src/components/Profile/WishlistCollection.tsx
+++ b/client/src/components/Profile/WishlistCollection.tsx
@@ -33,7 +33,10 @@ function WishlistCollection() {
 
   return (
     <>
-      <div>
+      <div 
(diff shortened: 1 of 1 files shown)


client/src/components/Profile/WishlistCollection.tsx
```diff
@@ -33,7 +33,10 @@ function WishlistCollection() {
 
   return (
     <>
-      <div>
+      <div
```
simonv3, line 36: probably a sign it needs to be a component that gets reused :P

LouisLouiseKay:
Yeah agreed, overall I've been meaning to create a padding component for a while so that we have a bunch of standard section paddings but I kept doing other stuff 🙃

PR #274, opened 2023-11-20
Code Clean Up

LouisLouiseKay:
This is mostly to clean up manually adjusted side paddings and font-size, created variables instead so it's easier to tweak global paddings.

Also did some overall adjustments to improve legibility on mobile.

### DIFF
--- a/client/src/components/Artist/Artist.tsx
+++ b/client/src/components/Artist/Artist.tsx
@@ -13,7 +13,9 @@ export const ArtistSection = styled.div`
   background: var(--mi-light-background-color);
   margin-bottom: 2rem;
   @media screen and (max-width: ${bp.medium}px) {
-    padding: 0.5rem !important;
+    padding-top: 0.5rem !important;
+    padding-bottom: 0.5rem !important;
+    padding: var(--mi-side-paddings-xsmall);
   }
 `;
 
--- a/client/src/components/ManageArtist/Manage.tsx
+++ b/client/src/components/ManageArtist/Manage.tsx
@@ -61,7 +61,9 @@ export const Manage: React.FC = () => {
             margin-top: 0 !important;
           }
           @media screen and (max-width: ${bp.medium}px) {
-            padding: .5rem;
+            padding: var(--mi-side-paddings-xsmall);
+            padding-top: .5rem;
+            padding-bottom: .5rem;
           }
 
         `}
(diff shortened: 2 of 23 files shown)


client/src/components/Artist/Artist.tsx
```diff
@@ -13,7 +13,9 @@ export const ArtistSection = styled.div`
   background: var(--mi-light-background-color);
   margin-bottom: 2rem;
   @media screen and (max-width: ${bp.medium}px) {
-    padding: 0.5rem !important;
+    padding-top: 0.5rem !important;
```
simonv3, line 16: Here you're overwriting `padding-top` and `padding-bottom` with the next line.

client/src/components/Artist/Artist.tsx
```diff
@@ -13,7 +13,9 @@ export const ArtistSection = styled.div`
...
+    padding-bottom: 0.5rem !important;
+    padding: var(--mi-side-paddings-xsmall);
```
simonv3, line 18:
```suggestion
    padding: var(--mi-side-paddings-xsmall);
```

Here you're overwriting `padding-top` and `padding-bottom` with the next line.

client/src/components/ManageArtist/Manage.tsx
```diff
@@ -61,7 +61,9 @@ export const Manage: React.FC = () => {
             margin-top: 0 !important;
           }
           @media screen and (max-width: ${bp.medium}px) {
-            padding: .5rem;
+            padding: var(--mi-side-paddings-xsmall);
+            padding-top: .5rem;
+            padding-bottom: .5rem;
```
simonv3, line 66: If the intention in the above ones is to have the padding-top and the padding-bottom overwrite the padding, then this would be the way to do it,

but maybe that's a sign that needs its own variable or component

client/src/components/ManageArtist/ManageArtistPosts.tsx
```diff
@@ -70,7 +70,8 @@ const ManageArtistPosts: React.FC<{}> = () => {
 
       @media screen and (max-width: ${bp.medium}px) {
         border-radius: 0;
-        padding: 1rem .5rem 0rem;
+        padding-top: 1rem;
+        padding: var(--mi-side-paddings-xsmall);
       `}
     >
       <div
```
simonv3, line 77: This can be replaced with `<HeaderDiv>`

PR #277, opened 2023-11-20
Some more adjustments to paddings and font-sizes for legibility.

LouisLouiseKay:
Adding font-size variables in different places to make text-sizes more consistent. Switched Release Page to a row of 4 albums. (Can be switched back if we want)

### DIFF
--- a/client/src/components/Home.tsx
+++ b/client/src/components/Home.tsx
@@ -131,6 +131,8 @@ function Home() {
                     line-height: 1rem;
                     padding: 1rem;
                     text-decoration: none;
+                    text-align: center;
+
                     &:hover {
                       text-decoration: underline;
                     }
@@ -154,6 +156,7 @@ function Home() {
                     line-height: 1rem;
                     padding: 1rem;
                     text-decoration: none;
+                    text-align: center;
                     &:hover {
                       text-decoration: underline;
                     }
--- a/client/src/components/ManageArtist/ManageArtistContainer.tsx
+++ b/client/src/components/ManageArtist/ManageArtistContainer.tsx
@@ -9,12 +9,7 @@ import { useArtistContext } from "state/ArtistContext";
 
 const Container = styled.div<{ artistBanner: boolean }>`
 
-
   width: 100%;
-
-  @media screen and (max-width: ${bp.medium}px) {
-    padding: var(--mi-side-paddings-xsmall);
-  }
 
   margin-top: calc(16vh);
 
@@ -22,11 +17,11 @@ const Container = styled.div<{ artistBanner: boolean }>`
   max-width: calc(1080px + 4rem);
 
   @media screen and (max-width: ${bp.medium}px) {
+    padding: var(--mi-side-paddings-xsmall);
     padding: 0rem !important;
     width: 100%;
-    // margin-top: 55px;
-    // margin-top: 30vw;
-    margin-top: 0;
+    margin-top: 0px;
+    ${(props) => (!props.artistBanner ? "margin-top: 0px;" : "")}
   }
 `;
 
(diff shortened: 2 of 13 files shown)


client/src/components/ManageArtist/ManageArtistContainer.tsx
```diff
@@ -35,7 +30,12 @@ export const ArtistPageWrapper: React.FC<{
   artistBanner: boolean;
 }> = ({ children, artistBanner }) => {
   return (
-    <Container artistBanner={artistBanner}>
+    <Container artistBanner={artistBanner}
+      className={css`
+        @media screen and (max-width: ${bp.medium}px) {
```
simonv3, line 35: This is empty, can it be deleted?

client/src/components/common/PageHeader.tsx
```diff
@@ -32,11 +32,11 @@ const PageHeader = () => {
             //* border-bottom: solid 4px; *//
             box-shadow: inset 1em -2em 0.8em -1.3em rgba(0, 0, 0, 0.4);
             @media screen and (max-width: ${bp.medium}px) {
+              position: absolute;
               ${user ? "margin-top: 0px;" : "height: calc(34vh);"}
               // --header-cover-sticky-height: 48px;
```
simonv3, line 37: Can comments be deleted?

client/src/components/ManageArtist/ManageArtistContainer.tsx
```diff
@@ -3,42 +3,43 @@ import React from "react";
...
   @media screen and (max-width: ${bp.medium}px) {
+    padding: var(--mi-side-paddings-xsmall);
     padding: 0rem !important;
     width: 100%;
-    // margin-top: 55px;
-    // margin-top: 30vw;
-    margin-top: 0;
+    ${(props) => (props.artistBanner ? "margin-top: 0px;" : "")}
+    ${(props) => (!props.userId ? "margin-top: 13vh;" : "")}
+    ${(props) => (!props.artistBanner ? "margin-top: 0px;" : "")}
```
simonv3, line 24: The `?` indicates a [ternary clause](https://stackoverflow.com/questions/392932/how-do-i-use-the-conditional-ternary-operator). 

```
condition ? do-the-true-thing : do-the-false-thing
```

So having a function call for both `!props.artistBanner` and `props.artistBanner` is redundant. You can instead do: 

```suggestion
    ${(props) => (props.artistBanner ? "margin-top: 0px;" : "margin-top: 0px;")}
    ${(props) => (!props.userId ? "margin-top: 13vh;" : "")}
```

Which is telling me that basically you're saying "if there is an artistBannder, do 0px margin, if there isn't an artistBanner, do 0px margin". Seems redundant? You should just have margin-top: 0px;

```suggestion
    margin-top: 0px;
    ${(props) => (!props.userId ? "margin-top: 13vh;" : "")}
```

client/src/components/common/ArtistHeaderSection.tsx
```diff
@@ -155,8 +167,13 @@ const ArtistHeaderSection: React.FC<{ artist: Artist; isManage?: boolean }> = ({
                 )}
               </div>
             </HeaderDiv>
-            <div className={css` filter: brightness(75%); opacity: .8;`}>
-            <MarkdownContent content={artist.bio}/>
+            <div
```
simonv3, line 170: If this is going to be the case for all `<MarkdownContent {...}` does it make sense to just have this set as the CSS in the `MarkdownContent` component?

client/src/components/Home.tsx
```diff
@@ -154,6 +156,7 @@ function Home() {
                     line-height: 1rem;
                     padding: 1rem;
                     text-decoration: none;
+                    text-align: center;
```
simonv3, line 159: It feels like these should get changed into re-usable CSS

client/src/components/common/ArtistHeaderSection.tsx
LouisLouiseKay: It's just for the artist bio! It could be a variable class though for muted text I think we don't have that in style yet.

PR #288, opened 2023-11-21
A Bunch of minor UI Fixes

LouisLouiseKay:
Adjusted Artist Header so that it works well with and without Avatar

Fixed Track Rows not displaying properly when hovered or when a track couldn't be played

### DIFF
--- a/client/src/components/common/ArtistHeaderSection.tsx
+++ b/client/src/components/common/ArtistHeaderSection.tsx
@@ -12,14 +12,12 @@ import { useTranslation } from "react-i18next";
 import HeaderDiv from "./HeaderDiv";
 
 const H1 = styled.h1`
-  font-size: 50px;
-  font-size: 2.6vmax;
-  line-height: 2.8vmax;
+  font-size: 45px;
+  line-height: 50px;
 
   @media screen and (max-width: ${bp.medium}px) {
-    font-size: 32px;
-    font-size: 3.2vmax;
-    line-height: 3.2vmax;
+    font-size: 27px;
+    line-height: 32px;
     padding-top: 0rem;
     padding-bottom: 0rem;
   }
@@ -36,7 +34,7 @@ const Header = styled.div`
   font-size: var(--mi-font-size-normal);
 
   p {
-    margin-bottom: 2rem;
+    // margin-bottom: 2rem;
   }
 
   @media screen and (max-width: ${bp.medium}px) {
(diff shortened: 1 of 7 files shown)


client/src/components/common/ArtistHeaderSection.tsx
```diff
@@ -12,14 +12,12 @@ import { useTranslation } from "react-i18next";
...
-  font-size: 2.6vmax;
-  line-height: 2.8vmax;
+  font-size: 45px;
+  line-height: 50px;
 
   @media screen and (max-width: ${bp.medium}px) {
-    font-size: 32px;
-    font-size: 3.2vmax;
-    line-height: 3.2vmax;
+    font-size: 27px;
```
simonv3, line 19: Feel like this should be `rem` or `em` not a pixel size. https://joshcollinsworth.com/blog/never-use-px-for-font-size

client/src/components/common/ArtistHeaderSection.tsx
```diff
@@ -141,7 +167,12 @@ const ArtistHeaderSection: React.FC<{ artist: Artist; isManage?: boolean }> = ({
                   )}
                 </div>
               </HeaderDiv>
-              <MarkdownContent content={artist.bio} />
+              <MarkdownContent
+                content={artist.bio}
+                className={css`
+                  markdownWrapper {
```
simonv3, line 173: this can be removed

client/src/components/common/Table.tsx
```diff
@@ -14,7 +14,7 @@ export const Table = styled.table`
 
       @media (prefers-color-scheme: dark) {
         background-color: ${(props) =>
-          colorShade(props.theme.colors.backgroundDark, -40)} !important;
+          colorShade(props.theme.colors.background, -40)} !important;
```
simonv3, line 17: Hmm, I feel like we should switch away from these variables entirely. Should be a ticket

simonv3 (dismissed):

client/src/components/common/ClickToPlayAlbum.tsx
```diff
@@ -0,0 +1,193 @@
+import styled from "@emotion/styled";
```
simonv3, line 1: This looks like it's mainly a copy of the functionality of the ClickToPlay component? I think we shouldn't duplicate components like that, if it's just the style that's different. It creates more places for bugs to exist that we have to squash in multiple places.

PR #290, opened 2023-11-22
Llk artistsectionavatar

LouisLouiseKay:
This is mainly improvements to artist header section to make sure it works elegantly with avatar, without avatar, with long band names, with short band names, in mobile view and in browser view.


client/src/components/ManageArtist/ManageTrackRow.tsx
```diff
@@ -96,6 +96,9 @@ const ManageTrackRow: React.FC<{
           overflow: hidden;
           whitespace: nowrap;
           text-overflow: ellipsis;
+          :hover {
```
simonv3, line 99: You don't have to do `&:hover` here for it to work? 

```suggestion
          &:hover {
```

client/src/components/Releases.tsx
```diff
@@ -24,21 +24,32 @@ const Releases = () => {
...
         className={css`
           display: flex;
           width: 100%;
           flex-direction: row;
           flex-wrap: wrap;
         `}
-      >    <div
-            className={css`
+      >
+        {" "}
```
simonv3, line 50: is this space intentional?

client/src/components/TrackGroup/TrackGroup.tsx
```diff
@@ -111,25 +112,39 @@ function TrackGroup() {
...
+              className={css`
+                display: flex;
+                margin-top: 1rem;
+                margin-bottom: 0.5rem;
+                align-items: center;
+                justify-content: flex-start;
+                align-items: center;
+              `}
+            >
+              {" "}
```
simonv3, line 126: Is this space intentional?

client/src/components/TrackGroup/TrackGroup.tsx
```diff
@@ -286,20 +301,43 @@ function TrackGroup() {
...
-        </div>
+          <div
+            className={css`
+              max-width: 70%;
+              margin: 1.25rem 0 1.25rem;
+              padding: 0.5rem 2rem 0.25rem 0rem;
+              border-right: 1px solid;
+              @media screen and (max-width: ${bp.small}px) {
+                max-width: 100%;
+                margin: 1.25rem 0 1.25rem;
```
simonv3, line 318: margin is the same in both no?

client/src/components/TrackGroup/TrackGroup.tsx
```diff
@@ -286,20 +301,43 @@ function TrackGroup() {
...
-            padding: 0.5rem 0.25rem;
-          `}
-        >
-          <MarkdownContent content={trackGroup.credits} />
+          <div
+            className={css`
+              margin: 1.25rem 0;
+              padding: 0.5rem 0.25rem 0.5rem 2rem;
+              font-size: var(--mi-font-size-small);
+              color: var(--mi-lighter-foreground-color);
```
simonv3, line 332: I feel like the `lighter` foreground colors should also be changed into transparency placed values

client/src/components/common/ArtistHeaderSection.tsx
```diff
@@ -95,55 +97,137 @@ const ArtistHeaderSection: React.FC<{ artist: Artist; isManage?: boolean }> = ({
...
+                <div
+                  className={css`
+                    ${artistAvatar ? "min-height: 50px;" : ""}
+                    display: flex;
+                    align-items: center;
+                    justify-content: space-between;
+                    width: 100%;
+                  `}
+                >
+                  <div>
```
simonv3, line 166: This shouldn't need a <div> wrapper? If it needs to be set to `display: block` I would set it on the H1 component

client/src/components/common/ArtistHeaderSection.tsx
```diff
@@ -95,55 +97,137 @@ const ArtistHeaderSection: React.FC<{ artist: Artist; isManage?: boolean }> = ({
...
+                    <H1
+                      className={css`
+                        @media screen and (max-width: ${bp.medium}px) {
```
simonv3, line 169: Can this just be added to the definition of `H1` above?

client/src/components/common/ArtistHeaderSection.tsx
```diff
@@ -95,55 +97,137 @@ const ArtistHeaderSection: React.FC<{ artist: Artist; isManage?: boolean }> = ({
...
+                        </Button>
+                      </Link>
+                    )}
+                  </div>
                 </div>
               </HeaderDiv>
-              <MarkdownContent content={artist.bio} />
+              <div
+                className={css`
+                  ${artistAvatar ? "display: none;" : ""}
```
simonv3, line 207: The ordering thing here doesn't seem to apply for not just using the ternary operator correctly for both of these

client/src/components/common/ArtistHeaderSection.tsx
```diff
@@ -95,55 +97,137 @@ const ArtistHeaderSection: React.FC<{ artist: Artist; isManage?: boolean }> = ({
...
+                  className={css`
+                    ${!artistAvatar ? "margin-bottom: .7rem !important" : ""}
+                  `}
+                />
+              </div>
             </div>
           </div>
+          <div
+            className={css`
+              ${!artistAvatar ? "display: none;" : ""}
```
simonv3, line 222: Ditto for ternary clause here

simonv3 (changes requested):

client/src/components/common/ArtistHeaderSection.tsx
LouisLouiseKay: It would be better but calling "!artistavatar" at the constant stage break things and I didn't know how to do it another way

client/src/components/common/ArtistHeaderSection.tsx
LouisLouiseKay: I think I fixed everything except that last one? What would I need to do here to make the code better? It's the only I found to make sure this didn't display when there's no avatar.

client/src/components/common/ArtistHeaderSection.tsx
LouisLouiseKay: Just to say I tried using props but couldn't manage to make it work, I'm sure the solution is obvious but I couldn't find it.

client/src/components/common/ArtistHeaderSection.tsx
simonv3: This and the next line could be turned into one statement

PR #302, opened 2023-11-25
Adjustments to code rework

LouisLouiseKay:
Just made the little things that the code clean up broke work again without adding much. 

Play button styling is better now.
Artist header section has better paddings and reordered buttons to avoid clutter.

### DIFF
--- a/client/src/components/Home.tsx
+++ b/client/src/components/Home.tsx
@@ -223,60 +223,53 @@ function Home() {
                 }
                 @media (prefers-color-scheme: dark) {
                   :hover {
-                    filter: brightness(110%);
+                    filter: brightness(140%);
                   }
                 }
               `}
             >
-              <div
+              <Box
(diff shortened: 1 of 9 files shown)


client/src/components/Home.tsx
```diff
@@ -223,75 +223,69 @@ function Home() {
...
                 `}
               >
-                <Box
-                  key={p.id}
+                <div
                   className={css`
-                    @media screen and (max-width: ${bp.medium}px) {
-                      width: 100% !important;
-                      padding: 2rem 2rem !important;
+                    display: flex;
```
simonv3, line 248: Feel like we need to institute a rule that if a CSS class is more than X lines it should be turned into a component. Wanna say 5 personally. Seriously impacts readability of the component, as well as confusing functionality with style.

simonv3 (approved):

PR #314, opened 2023-11-26
Llk artistsectionavatar

LouisLouiseKay:


client/src/components/Home.tsx
```diff
@@ -204,14 +204,27 @@ function Home() {
       {user && <Releases />}
       {user && posts.length > 0 && (
         <>
-          <h2
+          <h1
```
simonv3, line 207: This should be `h2` imo. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page

client/src/components/ManageArtist/ManageArtistContainer.tsx
```diff
@@ -10,7 +10,7 @@ import { useArtistContext } from "state/ArtistContext";
 
 const Container = styled.div<{ artistBanner: boolean; userId?: number }>`
   width: 100%;
-  margin-top: calc(16vh);
+  margin-top: calc(12vh);
```
simonv3, line 13: I feel like this is changing on every PR 😆

client/src/components/Profile/Collection/index.tsx
```diff
@@ -46,81 +46,20 @@ function Profile() {
             flex-direction: row;
             flex-wrap: wrap;
           `}
-        >    <div
```
simonv3, line 49: ❤️

client/src/components/common/TrackRow.tsx
```diff
@@ -174,16 +178,23 @@ const TrackRow: React.FC<{
...
             e.stopPropagation();
             navigator.clipboard.writeText(widgetUrl(track.id));
             snackbar("Copied track url", { type: "success" });
           }}
+          className={css`
+            :hover {
+              background: transparent !important;
+              opacity: 0.6;
+            }
+          `}
```
simonv3, line 194: Should this be a `iconOnly` button with `startIcon={<FiLink />}`?

simonv3 (approved):

client/src/components/Home.tsx
LouisLouiseKay: No problem, I'll have to refactor most header behaviors at some point to be more consistent across pages (I've started doing that already and this was part of it but I'll change what I've done with H1 to function with H2)

client/src/components/ManageArtist/ManageArtistContainer.tsx
LouisLouiseKay: Oh on this one it just changed because the header became transparent on artist page ! so it left us more margin to visualize more of the artist page at first glance.

simonv3 (approved):

PR #340, opened 2023-12-04
Modal Component improvement

LouisLouiseKay:
Modal component had several quirks that made using it frustrating 

- You can now close it anytime (header and close buttons are sticky)
- Height is now based on viewport which allows for a more consistent user experience across varying devices
- Small modal is slightly bigger to avoid it shrinking into a narrow column on medium bp
- scrollbar is now less proeminent in the box to avoid having up to 4 big scrollbars on the same window sometimes.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -147,6 +147,7 @@ function App() {
                       padding: var(--mi-side-paddings-xsmall);
                       padding-top: 0.5rem;
                       padding-bottom: 0.5rem;
+                      width: 100%;
                     }
                   }
                 `}
--- a/client/src/components/TrackGroup/PurchaseOrDownloadAlbumModal.tsx
+++ b/client/src/components/TrackGroup/PurchaseOrDownloadAlbumModal.tsx
@@ -94,7 +94,7 @@ const PurchaseOrDownloadAlbum: React.FC<{
                   }
                 `}
               >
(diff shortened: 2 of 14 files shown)


client/src/App.tsx
```diff
@@ -147,6 +147,7 @@ function App() {
                       padding: var(--mi-side-paddings-xsmall);
```
simonv3, line 147: I feel like if there's a div that has consistent styles in all the <Outlet> spaces then the styles should be defined on those `div`s and not here, and the <div> should be turned into its own re-used component.

LouisLouiseKay: You're right! That was actually something I wanted to do but didn't know how at the time but now that I've done the AlbumGrid Component I feel like it's fairly similar and I should be able to do it right?

client/src/components/TrackGroup/PurchaseOrDownloadAlbumModal.tsx
```diff
@@ -83,23 +83,39 @@ const PurchaseOrDownloadAlbum: React.FC<{
                 }
               `}
             >
-              <Button compact onClick={() => setIsPurchasingAlbum(true)}>
+              <Link
```
simonv3, line 86: Don't convert a Button to a Link

PR #346, opened 2023-12-05
Refactor manage artist detail, buy button and modal adjustments

LouisLouiseKay:
Lots of tweaks to : 

Manage artist detail (it now looks much more intentionally styled)
Buy button for adapted behaviors based on album page or if it's in artist page (+ mobile/browser changes)
Adjustments to modal to make it work better on mobile
Various minor fixes here and there

### DIFF
--- a/client/src/components/Header/Menu.tsx
+++ b/client/src/components/Header/Menu.tsx
@@ -119,7 +119,11 @@ const Menu: React.FC = (props) => {
               {t("collection")}
             </Button>
           </li>
-          <li>
+          <li
+            className={css`
+              border-bottom: solid 1px var(--mi-lighter-foreground-color);
+            `}
+          >
             <Button
               onClick={() => {
                 setIsMenuOpen(false);
@@ -151,7 +155,11 @@ const Menu: React.FC = (props) => {
               </li>
             );
           })}
-          <li>
+          <li
+            className={css`
+              border-top: solid 1px var(--mi-lighter-foreground-color);
+            `}
+          >
             <Button onClick={onLogOut} className={css``}>
               {t("logOut")}
             </Button>
@@ -178,7 +186,7 @@ const Menu: React.FC = (props) => {
       <li
         onClick={() => {
           setIsMenuOpen(false);
-          navigate("/about");
+          navigate("/pages/about");
         }}
       >
         <Button>{t("about")}</Button>
--- a/client/src/components/Header/Menu.tsx
+++ b/client/src/components/Header/Menu.tsx
@@ -66,6 +66,7 @@ const Menu: React.FC = (props) => {
           display: block;
           white-space: normal;
           color: var(--mi-black);
+          word-break: break-word;
           font-weight: normal;
           border-radius: 0;
           padding: 0.5rem;
--- a/client/src/components/ManageArtist/ArtistForm.tsx
+++ b/client/src/components/ManageArtist/ArtistForm.tsx
@@ -199,6 +199,7 @@ export const ArtistForm: React.FC<{
     >
       <div
         className={css`
+          margin-top: 0.5rem;
           label {
             display: block;
             font-size: 1.5rem;
--- a/client/src/components/ManageArtist/ArtistForm.tsx
+++ b/client/src/components/ManageArtist/ArtistForm.tsx
@@ -6,7 +6,7 @@ import { bp } from "../../constants";
 import api from "services/api";
 import { InputEl } from "../common/Input";
 import FormComponent from "components/common/FormComponent";
-import FormSectionWrapper from "components/common/FormSectionWrapper";
+import styled from "@emotion/styled";
 import TextArea from "components/common/TextArea";
 import { css } from "@emotion/css";
 import { useSnackbar } from "state/SnackbarContext";
@@ -22,6 +22,9 @@ export interface ShareableTrackgroup {
   creatorId: number;
   slug: string;
 }
+const FormSectionWrapper = styled.div`
+  margin-top: 2rem;
+`;
 
 type FormData = {
   name: string;
--- a/client/src/components/ManageArtist/ArtistForm.tsx
+++ b/client/src/components/ManageArtist/ArtistForm.tsx
@@ -197,96 +197,105 @@ export const ArtistForm: React.FC<{
       onClose={onClose}
       title={existing ? "Edit artist" : "Create an artist"}
     >
-      <FormProvider {...methods}>
-        <form onSubmit={handleSubmit(soSave)}>
-          <div
-            className={css`
-              display: flex;
-            `}
-          >
+      <div
+        className={css`
+          label {
+            font-size: 1.5rem;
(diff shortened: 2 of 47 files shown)


client/src/components/Header/Menu.tsx
```diff
@@ -118,7 +119,11 @@ const Menu: React.FC = (props) => {
               {t("collection")}
             </Button>
           </li>
-          <li>
+          <li
```
simonv3, line 122: Could be turned into a re-used component  if it's the same as below

client/src/components/ManageArtist/ArtistForm.tsx
```diff
@@ -243,28 +260,47 @@ export const ArtistForm: React.FC<{
               {t("displayName")}:{" "}
               <InputEl {...register("name", { required: true })} />
             </FormComponent>
-            {existing && <ArtistFormColors />}
+            <div
```
simonv3, line 263: Could be a re-used component

client/src/components/TrackGroup/PurchaseOrDownloadAlbumModal.tsx
```diff
@@ -22,7 +23,7 @@ const PurchaseOrDownloadAlbum: React.FC<{
   const { state: artistState } = useArtistContext();
 
   const userId = user?.id;
-
+  const { trackGroupId } = useParams();
```
simonv3, line 26: Why not use `const trackGroupId = trackGroup.id`?

client/src/components/TrackGroup/PurchaseOrDownloadAlbumModal.tsx
```diff
@@ -76,26 +77,40 @@ const PurchaseOrDownloadAlbum: React.FC<{
                 margin-top: 0rem;
               `}
             >
+              <Button
+                variant="outlined"
+                className={css`
+                  display: block !important;
+                  ${!trackGroupId ? "display: none !important;" : ""}
```
simonv3, line 84: When would this ever be non-existent? Is this essentially saying "don't show the purchase button on the group pages?" That feels like it should be defined by a variable to the component, not defined by the URL. 

```
<PurchaseOrDownloadAlbumModal
  trackGroup={trackGroup}
  showAsLink
  />
```

And then here simply have something like 
```
css`${showAsLink ? "" : ""}` 
```

Would be a lot less code re-use

simonv3 (changes requested):

client/src/components/TrackGroup/PurchaseOrDownloadAlbumModal.tsx
LouisLouiseKay: Ah this looks good, yeah the point was to not display the button style unless : 

1/ It's trackgroup AND It's browser view

client/src/components/TrackGroup/PurchaseOrDownloadAlbumModal.tsx
LouisLouiseKay: Don't know... I did this cause it worked.

client/src/components/Header/Menu.tsx
simonv3: Ah gotcha

client/src/components/ManageArtist/ArtistForm.tsx
simonv3: A styled-component looks like 

```
const Div = styled.div`
  margin-top: 2rem;
`
```

And then you're replace the next few lines with `<Div>`

client/src/components/TrackGroup/PurchaseOrDownloadAlbumModal.tsx
simonv3: `useParams` means it would only get a trackGroupId if it was defined in the URL. I'd use `trackGroup.id` and add a variable to the component (as described below)

client/src/components/ManageArtist/ArtistForm.tsx
```diff
@@ -20,6 +22,9 @@ export interface ShareableTrackgroup {
   creatorId: number;
   slug: string;
 }
+const FormSectionWrapper = styled.div`
```
simonv3, line 25: And now this begets the question, could this be replaced with `<FormComponent>`? The argument for is that it would standardize the way form elements look! :)

client/src/components/pages/FAQ.tsx
```diff
@@ -8,7 +9,14 @@ const FAQ: React.FC = () => {
   const { t } = useTranslation("translation", { keyPrefix: "faq" });
 
   return (
-    <MarkdownWrapper>
+    <MarkdownWrapper
+      className={css`
```
simonv3, line 13: This could be a re-used className! 

```
const pageMarkdownWrapper = css`
  ...
`
```

And then it could be used in FAQ, About, and CookiePolicy

client/src/components/ManageArtist/ArtistForm.tsx
```diff
@@ -255,44 +251,35 @@ export const ArtistForm: React.FC<{
...
 
-          <div
-            className={css`
-              margin-top: 1rem;
-            `}
-          >
+          <div>
             <FormComponent>
-              {t("displayName")}:{" "}
+              <h2>{t("displayName")} </h2>
```
simonv3, line 256: I don't think these should be headers. Should be `<label>` at best.

client/src/components/ManageArtist/UploadArtistImage.tsx
```diff
@@ -32,7 +32,7 @@ const UploadArtistImage: React.FC<{
         `}
       >
         <label>
-          <strong>{t(imageType)}</strong>
+          <h2>{t(imageType)}</h2>
```
simonv3, line 35: I don't know that this needs an other div if it's already a label.

client/src/components/common/Modal.tsx
simonv3, line 1: It looks like there's too many closing `}` here so that the css isn't formatting correctly.

client/src/components/ManageArtist/ArtistFormColors.tsx
```diff
@@ -8,17 +8,17 @@ const ArtistFormColors = () => {
   return (
     <>
       <label>
-        <strong>{t("customColors")}</strong>
+        <h2>{t("customColors")}</h2>
```
simonv3, line 11: I don't think this should be an `<h2>` but instead the `<label>` should be styled appropriately

client/src/components/ManageArtist/UploadImage.tsx
```diff
@@ -28,7 +29,10 @@ const UploadImage: React.FC<{
           <img
             src={existingImage}
             className={css`
-              max-width: 120px;
+              max-width: 150px;
```
simonv3, line 32: Let's refactor this in the future so that it gets re-used as below

simonv3 (approved):

PR #355, opened 2023-12-07
Fix home page post

LouisLouiseKay:
Posts were broken on the home page because they were displayed as markdown content.

Ideally that'd probably be better, we could probably create a system where only the "x first characters" are shown, which would allow us to have an excerpt and keep it as markdown.

For now the quick and dirty solution is to use nowrap and ellipsis so we can only have a one line excerpt in the post card.

### DIFF
--- a/client/src/components/Home.tsx
+++ b/client/src/components/Home.tsx
@@ -8,6 +8,7 @@ import Box from "./common/Box";
 import Logo from "./common/Logo";
 import { Trans, useTranslation } from "react-i18next";
 import Releases from "./Releases";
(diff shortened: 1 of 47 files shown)


client/src/components/Home.tsx
```diff
@@ -349,7 +348,8 @@ function Home() {
                         color: var(--mi-normal-foreground-color);
                       `}
                     >
-                      <MarkdownContent content={p.content} />
+                      {/* <MarkdownContent content={p.content} />*/}
+                      {p.content}
```
simonv3, line 352: isn't this showing it as unstructured markdown?

PR #373, opened 2023-12-17
Hide header on scrolldown, show on scrollup

LouisLouiseKay:
It's all in the title really

### DIFF
--- a/client/src/components/Header/Header.tsx
+++ b/client/src/components/Header/Header.tsx
@@ -10,8 +10,93 @@ import Menu from "./Menu";
 import Logo from "components/common/Logo";
 import DropdownMenu from "components/common/DropdownMenu";
 import { ImMenu } from "react-icons/im";
+import styled from "@emotion/styled";
+import { useState, useEffect } from "react";
+
+const HeaderWrapper = styled.div<{
+  artistBanner: boolean;
+  artistId: boolean;
+  show: boolean;
+  trackGroupId: boolean;
+}>`
+  position: sticky;
+  width: 100%;
+  z-index: 999;
+  transition: top 0.4s ease-out;
+
+  ${(props) =>
+    props.artistBanner && props.artistId
+      ? "background: transparent; box-shadow: 0px 1px 10px rgba(0, 0, 0, 0);"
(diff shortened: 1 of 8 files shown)


client/src/components/Home.tsx
```diff
@@ -241,73 +243,85 @@ function Home() {
...
-                @media screen and (max-width: ${bp.xlarge}px) {
-                  padding: var(--mi-side-paddings-xsmall);
-                }
+          <div
+            className={css`
+              background-color: var(--mi-light-background-color);
 
-                @media screen and (max-width: ${bp.medium}px) {
-                  flex-direction: column;
+              a {
```
simonv3, line 260: are we sure we want to apply this to all links inside this div container?

client/src/components/Header/Header.tsx
```diff
@@ -25,50 +110,12 @@ const Header = () => {
...
-          background-color: var(--mi-black);
-          ${trackGroupId ? "background-color: transparent;" : ""}
-          ${artistBanner ? "background-color: transparent;" : ""}
-
-          @media screen and (max-width: ${bp.medium}px) {
-            border-bottom: 1px solid transparent;
-          }
-        }
-      `}
+    <HeaderWrapper
```
simonv3, line 113: Appreciate this!

simonv3 (approved):

client/src/components/Home.tsx
LouisLouiseKay: Yeah cause it's all stuff that you click on anyway. Like links were broken for stuff where there's a link inside a sentence, but for article titles, album titles etc. There's just no need to accentuate them more (they all have underline on hover anyway). It's just important to keep links colored in markdown, blogposts etc. Wherever it's not clear that some things in the text/UI are clickable.

simonv3: Then shouldn't this be universal? In styles.ts?

LouisLouiseKay: It's just for home page, it's because basically on home page it's ALL links. Everything is clickable so it's the opposite for accessibility it's worse if it's all colored. But then as soon as it's not all link it goes back to default rule - (links have a different color + underlined on hover)

LouisLouiseKay: (see footer where links are displayed normally)

PR #374, opened 2023-12-17
Minor UI fixes

LouisLouiseKay:
Lots of UI fixes in a bunch of places, things that were a little broken here and there.


client/src/components/common/Logo.tsx
```diff
@@ -1,9 +1,12 @@
 import React from "react";
 import { css } from "@emotion/css";
+import { useParams } from "react-router-dom";
 
-export const Logo: React.FC<{ noWordmark?: boolean }> = ({
+export const Logo: React.FC<{ noWordmark?: boolean; artistId: boolean }> = ({
   noWordmark = false,
+  artistId: boolean,
 }) => {
+  const { artistId } = useParams();
```
simonv3, line 9: This doesn't cause an error defining the same variable twice? I think in settings like this it's useful to name booleans with a `is` prefix. Like `isArtist`. It makes it much clearer what the variable is containing. Right now it looks like the variable contains an artist id.

client/src/components/Header/HeaderSearch.tsx
```diff
@@ -93,6 +93,11 @@ const HeaderSearch: React.FC = () => {
               background: var(--mi-white);
               width: 100%;
               z-index: 12;
+              word-break: break-word;
+                color: var(--mi-black) !important;
```
simonv3, line 97: This loks like prettier didn't run which probably means that the css isn't formatted correctly, which probably means it's not displaying some things

client/src/components/Header/Header.tsx
```diff
@@ -138,7 +138,7 @@ const Header = () => {
                 }
               `}
             >
-              <Logo />
+              <Logo artistId />
```
simonv3, line 141: See note about this being a boolean below

client/src/components/common/Logo.tsx
LouisLouiseKay: Ah yeah, sorry I put that in because it was the proposed fix I think! I actually didn't even need to put, it worked fine without it, but it threw an error that's why I added it in.

client/src/components/Header/HeaderSearch.tsx
LouisLouiseKay: Oh yeah weird easy fix

PR #377, opened 2023-12-18
Fix to the header scroll behavior on album page.

LouisLouiseKay:
It didn't display properly on mobile because it assumed it had a banner.

### DIFF
--- a/client/src/components/Header/Header.tsx
+++ b/client/src/components/Header/Header.tsx
@@ -16,7 +16,7 @@ import { useState, useEffect } from "react";
 const HeaderWrapper = styled.div<{
   artistBanner: boolean;
   artistId: boolean;
-  show: boolean;
+  show: string;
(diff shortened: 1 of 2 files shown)


client/src/components/Header/Header.tsx
```diff
@@ -73,29 +75,31 @@ const HeaderWrapper = styled.div<{
...
-    };
-    window.addEventListener("scroll", controlNavbar);
-
-    // cleanup function
-    return () => {
-      window.removeEventListener("scroll", controlNavbar);
-    };
-  }, [lastScrollY]);
+  function useShow() {
+    const [Show, setShow] = useState("{}");
```
simonv3, line 79: This should be lowercase 

```suggestion
    const [show, setShow] = useState("");
```

Also you'll likely not want curly braces right?

simonv3 (approved):

PR #387, opened 2023-12-19
Widget simple version

LouisLouiseKay:
A version that "just works" although it removes the capacity to use Mirlo player when widget is on Mirlo (which doesn't work currently anyway)


simonv3:
I'd rather fix the original functionality than merge this honestly :/

LouisLouiseKay:
Yeah no problem! It was just that it's broken right now but that's fine.

PR #389, opened 2023-12-20
Llk small fixes

LouisLouiseKay:


client/src/components/Post/index.tsx
```diff
@@ -17,9 +17,42 @@ export const pageMarkdownWrapper = css`
   margin: auto;
   padding: var(--mi-side-paddings-xsmall);
   font-size: 18px;
+  line-height: 1.7rem;
+
+  h1 {
```
simonv3, line 22: Not for now but I wonder if we can add automatic classes to post html components so that we don't have to do `!important` things

simonv3 (approved):

PR #453, opened 2024-01-03
Adjustments new artist page

LouisLouiseKay:
Mostly adjustments to how the artist page is now setup with the tabs.

A new "big" button variant to create more differences between different roles in different contexts.

### DIFF
--- a/client/src/components/common/ArtistHeaderSection.tsx
+++ b/client/src/components/common/ArtistHeaderSection.tsx
@@ -6,7 +6,6 @@ import styled from "@emotion/styled";
 import FollowArtist from "./FollowArtist";
 import { useGlobalStateContext } from "state/GlobalState";
 import Button from "./Button";
-import { FaPen } from "react-icons/fa";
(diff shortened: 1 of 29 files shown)


client/src/components/common/ArtistHeaderSection.tsx
```diff
@@ -176,48 +175,42 @@ const ArtistHeaderSection: React.FC<{ artist: Artist; isManage?: boolean }> = ({
...
                       {!isManage && <FollowArtist artistId={artist.id} />}
                       {!isManage && user?.id === artist.userId && (
                         <Link to={`/manage/artists/${artist.id}`}>
-                          <Button
-                            compact
-                            transparent
-                            type="button"
-                            startIcon={<FaPen />}
-                          >
+                          <Button compact variant="big" type="button">
```
simonv3, line 195: off the cuff does it make sense for a button to be both compact and big?

simonv3 (approved):

PR #464, opened 2024-01-05
Fixed bottom padding

LouisLouiseKay:


simonv3:
Build fails due to isPlaying not being used anymore likly

LouisLouiseKay:
> Build fails due to isPlaying not being used anymore likly

(Which I'd know if I'd read the error message)

PR #481, opened 2024-01-08
Minor fixes

LouisLouiseKay:
Two minor fixes in mobile view, trackgroup card didn't keep aspect ratio for album images and the tables for supporters were overflowing the main screen.

### DIFF
--- a/client/src/components/ManageArtist/TrackGroupCard.tsx
+++ b/client/src/components/ManageArtist/TrackGroupCard.tsx
@@ -115,8 +115,6 @@ const TrackGroupCard: React.FC<{
               }
             }
           }
-          @media screen and (max-width: ${bp.small}px) {
-          }
         `}
       >
         <div>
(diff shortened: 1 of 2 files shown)


client/src/components/ManageArtist/TrackGroupCard.tsx
```diff
@@ -107,6 +115,8 @@ const TrackGroupCard: React.FC<{
               }
             }
           }
+          @media screen and (max-width: ${bp.small}px) {
```
simonv3, line 118: Delete

client/src/components/common/Table.tsx
```diff
@@ -29,6 +29,23 @@ export const Table = styled.table`
       padding: 0.25rem 0.5rem;
     }
   }
+  @media screen and (max-width: 576px) {
```
simonv3, line 32: Should this be a screen breakpoint?

LouisLouiseKay: Yeah I put it that way cause it's also a number instead of a bp above so I wondered if there was a reason to not import more stuff just for that and kept the same logic.

PR #561, opened 2024-02-08
Reworked widget

LouisLouiseKay:
Fixes to widget so that it displays well in mirlo posts, plus overhaul of visual styling of the widgets


client/src/components/Player/PlayingTrackDetails.tsx
```diff
@@ -20,6 +20,10 @@ const PlayingTrackDetails: React.FC<{ currentTrack: Track }> = ({
         margin-bottom: 0.3rem;
         padding-top: 0.5rem;
 
+        a {
+          text-decoration: none;
```
simonv3, line 24: I think we should keep the underline for accessibility reasons.

LouisLouiseKay: Wonder if there isn't a better way? I just find them extremely visually distracting and to me it impairs readibility more than it helps. I'm not familiar. On small isolated links I find them ok, or text within a text where you need to know what's clickable. On entire strings of texts, I really don't find them all that good, since the whole card is clickable anyway in the poscard.

Issue #629, opened 2024-03-25
Navigating to an artist page and quickly back to the homepage breaks homepage links


fennifith:
Hey! I've come across a slight bug while clicking around the site: it seems like navigating to an artist/album page, and then quickly navigating back to the homepage, will result in all links on the homepage pointing to that artist.

To reproduce:
- Optionally: Enable network throttling in DevTools (makes this easier to reproduce)
- Go to the mirlo.space homepage
- Click on an artist under "recent releases"
- Before the artist page loads, press the browser back button to go back home
- Observe that all links in "Recent releases" now point to the artist you just visited
  * i.e. after visiting `<artist1>`, instead of seeing a link to `/<artist2>/release/<artist2's album>`, the URL is `/<artist1>/release/<artist2's album>` - which leads to a 404

## Technical Details

This release section is rendered using `<ArtistTrackGroup>`, which will use the artist ID from `useArtistContext()` if provided: https://github.com/funmusicplace/mirlo/blob/main/client/src/components/Artist/ArtistTrackGroup.tsx#L67

I believe the `refreshArtist()` call in [ArtistContext.tsx](https://github.com/funmusicplace/mirlo/blob/main/client/src/state/ArtistContext.tsx#L106) may be causing this with the following sequence of events:

- The artist page is visited, and `refreshArtist()` is invoked - starting a `fetch()` call
- The browser navigates back to the homepage
- The artist context is correctly cleared when the homepage loads
- The fetch call from `refreshArtist()` finishes, which then calls `dispatch()` with the previously-visited artist

## Solutions

This *could* be fixed by providing an [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal) when `fetch()` is called, and aborting the fetch call when the requesting component is unmounted - e.g. in the [cleanup function of `useEffect`](https://react.dev/reference/react/useEffect#useeffect)

However, I'm wondering if y'all have thought about using [TanStack Query](https://tanstack.com/query/latest)? It effectively handles all of these edge cases (and more) out of the box - and I've seen it widely used for that reason.

I'd be happy to either take up fixing this with `AbortController` or providing a proof-of-concept with TanStack Query in a PR, or both - let me know!

simonv3:
Thank you for the really thorough bug report and also proposed solution! You have to be _really_ fast to replicate this hah. I'll take a look at your proposed fix. 

I'm down to try out a TanStack Query PR too if that's something you're interested in doing a punt for.

[...]

PR #630, opened 2024-03-26
Add AbortController to avoid fetch calls returning after page navigation

fennifith:
This appears to fix the issue described in #629 - although I'm not sure if the same behavior occurs on any other routes.


simonv3:
Thank you!

Issue #655, opened 2024-04-07
Blog post "View live" url doesn't redirect to post under artist url


simonv3:
/post/:id should be /:artistSlug/posts/:id

simonv3:
Ah, it looks like this should also happen on the general posts view https://mirlo.space/manage/artists/:artistId/posts @ericburns sorry!!

simonv3:
Thanks @ericburns !!

PR #658, opened 2024-04-08
#655 - Fix view live post URL for Artist

ericburns:
While editing a post on `/manage/artists/1/post/1`, the URL for the `View live` button previously went to `/post/:id`. It has been changed to link to `/:artistSlug/posts/:id`.

NOTE: I noticed that the manage posts page for artists (`/manage/artists/1/posts`) has the same issue. Should that be updated as well?

### DIFF
--- a/client/src/components/ManageArtist/ManagePost.tsx
+++ b/client/src/components/ManageArtist/ManagePost.tsx
@@ -75,7 +75,7 @@ const ManagePost: React.FC<{}> = () => {
                 align-items: center;
               `}
             >
-              <Link to={`/post/${post.id}`}>
+              <Link to={`/${artist?.urlSlug}/posts/${post.id}`}>
                 <Button type="button">{t("viewLive")}</Button>
               </Link>
             </div>


simonv3:
Thank you!

> NOTE: I noticed that the manage posts page for artists (/manage/artists/1/posts) has the same issue. Should that be updated as well?

This is less important I think since that link is less likely to get shared (it only works for the logged in artist).

client/src/components/ManageArtist/ManagePost.tsx
```diff
@@ -75,7 +75,7 @@ const ManagePost: React.FC<{}> = () => {
                 align-items: center;
               `}
             >
-              <Link to={`/post/${post.id}`}>
+              <Link to={`/${artist?.urlSlug}/posts/${post.id}`}>
```
simonv3, line 78: There's a helper function called `getPostURLReference` that would make it so that this is consistent, but it might need a little rewrite to work with a potentially missing artist. You can see it in ArtistPosts.tsx

simonv3 (approved):
Non-blocking comment

client/src/components/ManageArtist/ManagePost.tsx
ericburns: Oh yeah, thanks for pointing that out! Just took a peek at the code and it looks like `getPostURLReference` is taking into account a potentially missing artist. Since this is merged, I can make a new PR to swap to that.

Issue #471, opened 2024-01-06
Update Sharp NPM package


Chickensoupwithrice:
@simonv3 asked if I'd be interested in updating the Sharp package. 

they mentioned It'll also mean updating the node version of the docker containers in Dockerfile to a more recent one, and likely also updating the minio package version as an effect of that

Creating this issue to track that

Chickensoupwithrice:
I can't assign myself to issues, otherwise I would've

simonv3:
> it's been giving me more trouble than i anticipated. the NODE_VERSION on newer alpine docker images seems to clash with sharp's expectations

PR #681, opened 2024-04-19
Migrate client to Vite

### LINKED ISSUE #52: Switch off of create-react-app and to something like vite
CRA is no longer maintained. We should update to something like vite. 

https://github.com/reactjs/react.dev/pull/5487#issuecomment-1409720741

fennifith:
This migrates the client app from create-react-app to Vite - which fixes #52!

- Renames env variables from `process.env.REACT_APP_*` to `import.meta.env.VITE_*`.
- Implements code splitting with `async import()`s in `routes.tsx`
  * This mainly reduces the bundle size from the current 3MBish JS bundle into many smaller files, which are fetched dynamically for each route
  * This is also used separately in `AudioWrapper.tsx` to load `@gumlet/react-hls-player` only when needed (as that library is quite big on its own)
- Adds `type: "module"` to package.json
- Migrates from jest to vitest, replaces the placeholder `App.test.tsx` with a barebones (but working) `index.test.tsx`
- Updates the node version from `16.19.0` to `18.20.1` (minimum required by Vite/Vitest)

I've also added `@vite/plugin-legacy` as mentioned in #52 - however, this does not add any polyfills that might be needed, nor have I tested the build result on any older browser versions. I've left out the browserslist plugin, as [the legacy plugin appears to provide its behavior by default](https://github.com/vitejs/vite/tree/main/packages/plugin-legacy#targets).

There is *some* risk that the node upgrade breaks things in the API, although this should only affect dev environments (I haven't touched the configured docker image). I've done some poking around to test things, but nothing thorough.


.nvmrc
```diff
@@ -1 +1 @@
-16.19.0
\ No newline at end of file
+18.20.1
```
fennifith, line 1: Considering the repo node version is being updated, it'd be good to also update the API docker image to [18-alpine3.19](https://hub.docker.com/layers/library/node/18-alpine3.19/images/sha256-05583a00b42e064e4f0bb05fb2392133aa2a8728f50d16bd95ccc705ddf96bc9?context=explore) (leaving the dev version mismatched with the deployment could cause confusing/hidden behavior if the node version breaks anything).

Is there anything else that should be updated when changing this?

client/package.json
```diff
@@ -42,25 +35,16 @@
...
-  },
-  "eslintConfig": {
-    "extends": [
-      "react-app",
-      "react-app/jest"
-    ]
+    "start": "vite",
+    "build": "tsc --noEmit && vite build",
+    "preview": "vite preview",
+    "test": "tsc --noEmit && vitest"
```
fennifith, line 47: Question: Would it make sense to add a Github Action for client tests (as well as `tsc`) alongside the API test workflow? (will make a separate PR for this if so)

.nvmrc
simonv3: Hmm, is updating the entire repo node version required? 

There is (was?) an issue with newer node versions and sharp, the image processor that we use, see here https://github.com/funmusicplace/mirlo/issues/471

client/package.json
simonv3: Yes! Though I think my preference personally is focusing on writing e2e tests with something like cypress

client/src/components/TrackGroup/TrackGroupEmbed.tsx
```diff
@@ -67,7 +67,7 @@ const TrackGroupEmbed: React.FC<{
             onClick={() => {
               if (trackGroup.artist) {
                 navigator.clipboard.writeText(
-                  process.env.REACT_APP_CLIENT_DOMAIN +
+                  import.meta.env.VITE_CLIENT_DOMAIN +
```
simonv3, line 70: Just wondering out loud: would it be better to ultimately replace these with something like window.location? Is that naive?

client/src/routes.tsx
```diff
@@ -1,204 +1,299 @@
...
   {
     path: "/",
     element: <App />,
     errorElement: <ErrorPage />,
     children: [
-      { path: "", element: <Home /> },
-      { path: "pages/about", element: <About /> },
+      {
+        path: "",
+        async lazy() {
```
simonv3, line 30: Do you know how this will play with pre-renderers?

simonv3 (commented):
Thank you! Some questions in the comments

.nvmrc
fennifith: Yeah - it looks like the unit test run on this PR might've run into a similar issue.

I've done some poking at the docker images locally - it appears that the `volumes: ./:/var/www/api` mount is the cause, which replaces the `node_modules` that should be populated by `yarn install` on the container (hence the docker image does not have the right version of sharp installed). That said, while removing  this has fixed the sharp error, I've run into a few more issues as a result - so I'll see where I get with debugging that.

It shouldn't be required to update the entire repo's node version - so that might work as a temporary fix if I don't get anywhere. That said, [Node 16 is nearing the end of its maintenance status](https://nodejs.org/en/about/previous-releases) so I think this upgrade will still be needed at some point.

client/src/components/TrackGroup/TrackGroupEmbed.tsx
fennifith: That would make sense to me! I've found the [`URL(url, base)`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) constructor useful for similar logic in the past, as it ensures the result is a valid link - so this could be `new URL(getReleaseUrl(...), window.location.origin).toString()`

client/src/routes.tsx
fennifith: Nope! Does Mirlo currently use SSR? (I didn't think it did, but could be wrong)

[According to the react-router docs](https://reactrouter.com/en/main/guides/ssr#lazy-routes) it seems like lazy functions are evaluated for SSR, but might cause some issues when hydrating on the client. This could be fixed by manually awaiting the correct route functions on initial load (as shown in their code sample) - or by using the full [Remix](https://remix.run) framework, which is their suggestion.

I'm not sure how much would be needed to migrate to Remix - from their docs it looks like it's just adding a Vite plugin and then moving the routes into file-based route modules, but there could be other differences.

.nvmrc
simonv3: > it appears that the volumes: ./:/var/www/api mount is the cause, which replaces the node_modules that should be populated by yarn install on the container (hence the docker image does not have the right version of sharp installed)

It's entirely possible I don't know how docker works but do we not need to mount our code inside the docker container somehow? Maybe I'm misunderstanding. 

Either way, if you get it working I'll be grateful! Docker = not my forte.

client/src/routes.tsx
simonv3: > Nope! Does Mirlo currently use SSR? (I didn't think it did, but could be wrong)

We don't, but we use netlify's [prerendering](https://docs.netlify.com/site-deploys/post-processing/prerendering/) for places that don't load the full page -- ie social media, discord, etc

It's fairly easy to test though, we can use the netlify preview to test it in discord

.nvmrc
fennifith: Yeah! When the docker *image* is built, the `COPY . .` command will place every file from the build location into the image (aside from those in `.dockerignore` - which includes `node_modules`).

When the docker *container* is created, it uses the most recent docker image build, and mounts any volumes that are specified.

Since you're using nodemon, it's seems like these volume mounts are intended to sync changes on the host with the container for "live reloading" without rebuilding the image. With the `node_modules` change, this *could* still be done to a limited extent - but any changes to `package.json` requiring another `yarn install` would now need the image to be rebuilt.

Are these dockerfiles (and nodemon) also being used in production? If so, it might make sense to separate these configs into a "dev" config that uses nodemon, and a "deploy" that does not - which might simplify this config a little.

simonv3: > Are these dockerfiles (and nodemon) also being used in production? If so, it might make sense to separate these configs into a "dev" config that uses nodemon, and a "deploy" that does not - which might simplify this config a little.

They're not used on production (though ideally they would be I think), production runs on render, which uses the render.yaml. 

I think it'd be great to have a docker file that is able to be set up in a production environment.

fennifith: Hey! So I have a follow-up question to this - I've noticed that render.yaml has an `env: static` service that hosts the client build, but the actual website seems to be on Netlify. Does that mean the render web service is unused?

docker-compose.yml
fennifith, line 1: I've made quite a few changes to these configs - wanted to summarize them a bit:
- To avoid mounting volumes from the host for hot reloading, I've added [`docker compose watch`](https://docs.docker.com/compose/file-watch/) configuration - which instead rebuilds & restarts the entire container when changes occur.
  * This is a bit limited, but fairly quick thanks to some caching in the Dockerfile - it doesn't need to run a full `yarn install` every time.
  * Otherwise, `yarn api:watch` can still be used to run nodemon *outside of the container*, although it needs a bit of environment configuration.

- I've moved containers for the integration tests into a `test` profile (in `docker-compose.test.yml`), rather than merging properties onto the original config. This means that `docker compose run test` can be used while `docker compose watch` is running, and each will start their own containers without conflicting.

- These are newer docker features, which I'm assuming the V1 `docker-compose` script does not support. That said, [Compose V1 has stopped receiving updates, and V2 seems to be widely available](https://docs.docker.com/compose/migrate/), so I'm hoping this is okay to drop support for.

- To get the newer versions of `sharp` to work, I needed to upgrade to Yarn v4. This can be done using [`corepack enable`](https://nodejs.org/api/corepack.html), which will automatically use the `"packageManager"` property in the closest package.json.
  * Render [doesn't support newer Yarn versions out of the box](https://feedback.render.com/features/p/add-support-for-yarn-2), but I've got it working with `SKIP_INSTALL_DEPS=true` and running `corepack enable && yarn install --immutable` as part of the buildCommand.
  * This also caused some problems with Prisma, which [has some known issues with Yarn](https://github.com/prisma/prisma/issues/8765) - for which I've added a few workarounds. Instead of writing to `node_modules/.prisma`, it now generates its client in `prisma/__generated__`, which I've configured as a Yarn workspace (so that imports can be aliased as `@mirlo/prisma` and `@mirlo/prisma/client` - hence the refactoring).
   * I also configured the `client/` package as another workspace. This doesn't impact much, but it means that `yarn install` sets up both at once, and scripts such as `yarn client:start` can be used from anywhere in the project.

.nvmrc
simonv3: It is yeah. We switched to netlify for static stuff. It's currently disabled in render itself. We could remove it, but I feel like there's an argument for leaving it if anyone wants to use Render to spin things up themselves with minimal tweaking (a kind of "install with render" button)

test/README.md
```diff
@@ -1,13 +1,13 @@
...
-docker-compose -f docker-compose.yml -f test.docker-compose.yml up
+docker compose run test
 ```
 
-And then run the tests:
+To clean up these containers afterwards, the `--profile=test` argument needs to be passed to docker:
 
 ```
-docker exec -it blackbird-api yarn test
+docker compose --profile=test down
```
simonv3, line 12: there's no way to do this automatically?

simonv3 (approved):
This is a huge change! Thank you. This looks fine to me but I'd like to pull it down locally and see how the docker stuff takes. 

I'll also have to change some things on netlify. Unfortunately we don't have a stage environment set up (due to cost mainly) so I wanna try to think about how to deploy this in case something goes wrong (only worried about prisma changes not taking due to caching issues I think?) 

I feel like this felt like one of those "cascading changes uncovered requiring doing this other thing" but maybe in the future we could strategize in how to make the code base a bit more conducive to changing things more incrementally? I don't know if you have thoughts about that?

test/README.md
fennifith: There is `docker compose --profile=test up --exit-code-from=test`, although that stops *every* container after exiting, not just the containers in `--profile=test`. Meanwhile, `docker compose run --rm test` is another option, but it only removes the `test` container and still leaves redis/postgres/minio up - so I haven't found a perfect solution.

PR #696, opened 2024-04-30
Flatten buttons wrapped with links into a single element

fennifith:
Many links on the site use a  `<Link> <Button /> </Link>` structure, which renders a styled `<button>` element wrapped with an `<a href="...">` from react-router.

When using keyboard navigation, the inner `<button>` element is also focusable and announced to screen readers, even though it does not have any effect when used.

This PR flattens these components into one `<ButtonLink>`, which renders a single `<a>` element with the button styling.


simonv3 (approved):

simonv3:
lgtm! thank you.

PR #699, opened 2024-05-02
Releases list a11y

fennifith:
This adds a few small accessibility improvements to the releases list and related components:

- Places the releases list in a `<ul>`, with `role="list"` and `aria-labelledby` pointing to the Releases header
- Restructures the `<ClickToPlay>` component so that:
  * Action buttons become visible when focused
  * Action buttons are persistently displayed on touchscreens
  * The invisible `position: absolute` album cover link is hidden from the accessibility tree (this is duplicated by the album link)
  * "Go to album" text is also hidden from the accessibility tree (this is instead placed on the album link via `aria-label`)
- Added a fallback "Untitled" album name for albums that do not have a title.
  I did this to ensure that the album link is visible when tabbed over - but styled differently to distinguish it from an actual album name. There might be other options here - e.g. reusing the text "Go to album" instead of "Untitled"?
- Renamed the `<Button>` `role=` property to `buttonRole=` to avoid clashing with the `role=` HTML attribute
- Added `aria-hidden` to decorative button icons, and added an `aria-label` to the Play / Pause / Wishlist buttons


simonv3 (approved):

simonv3:
Thank you!

simonv3:
For this issue: 

> Added a fallback "Untitled" album name for albums that do not have a title.

From the album art for some of these I think people are also just not adding album titles even though a title exists. Which is a problem on another issue. Since I do think it should be possible to release an untitled album, I think we'll have to think through the create album flow a bit too

PR #714, opened 2024-05-06
Post card a11y and excerpt changes

fennifith:
This PR makes a few changes to post cards:
- Removes the wrapping `<a>` tag, and replaces it with `useLinkContainer` to handle click events
  * These mimic the behavior of a link for ctrl+click, middle click, and other mouse interactions, without actually involving an `<a>` element (which otherwise involves working with z-index, and prevents text selection from working as expected)
  * This is mostly reused from [unicorn-utterances/href-container-script.ts](https://github.com/unicorn-utterances/unicorn-utterances/blob/main/src/utils/href-container-script.ts)
- Moved some shared structure between the artist "Updates" page and homepage into the `<PostCard>` component
  * Consequently, the styling of the post cards in these locations is now the same.
- Added a `getHtmlExcerpt` utility to roughly parse out any text content from the post HTML
  * This means that the post excerpts only contain a text summary of the first 8 lines/paragraphs of a post.
  * Previously, tabbable/interactive nodes and headings were included in this preview, which created a lot of items to tab through for longer posts when the entire content is in the DOM.
  * However, this does prevent any formatting (bold, italics, blockquotes, images, etc.) from appearing in the post cards.

And on heading elements:
- Changed section headings to h2 so that the homepage heading levels are in sequential order
- Added `.h1`-`.h6` classes in addition to the element styles that can be used to visually set the heading size without changing the element

### DIFF
--- a/client/src/components/Artist/ArtistSupport.tsx
+++ b/client/src/components/Artist/ArtistSupport.tsx
@@ -14,7 +14,7 @@ import { useAuthContext } from "state/AuthContext";
 import { useQueryClient } from "@tanstack/react-query";
 import { QUERY_KEY_AUTH, queryKeyIncludes } from "queries/queryKeys";
 
-export const PostGrid = styled.div<{}>`
+const ArtistSupportGrid = styled.div`
   display: grid;
   grid-template-columns: repeat(3, 31.6%);
   gap: 4% 2.5%;
@@ -143,11 +143,11 @@ const ArtistSupport: React.FC = () => {
           {t("noSubscriptionTiersYet")}
         </Box>
       )}
-      <PostGrid>
+      <ArtistSupportGrid>
         {artist.subscriptionTiers
           ?.filter((p) => !p.isDefaultTier)
           .map((p) => <ArtistSupportBox key={p.id} subscriptionTier={p} />)}
-      </PostGrid>
+      </ArtistSupportGrid>
     </>
   );
 };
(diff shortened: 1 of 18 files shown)


client/src/components/Artist/ArtistSupport.tsx
```diff
@@ -9,11 +10,27 @@ import { bp } from "../../constants";
 import FollowArtist from "components/common/FollowArtist";
 import SpaceBetweenDiv from "components/common/SpaceBetweenDiv";
 import { useArtistContext } from "state/ArtistContext";
-import { PostGrid } from "./ArtistPosts";
 import { useAuthContext } from "state/AuthContext";
 import { useQueryClient } from "@tanstack/react-query";
 import { QUERY_KEY_AUTH, queryKeyIncludes } from "queries/queryKeys";
 
+export const PostGrid = styled.div<{}>`
```
simonv3, line 17: Should this get imported from Post/PostGrid?

simonv3 (approved):
Looks good! I like how you handled the except getting. One question about re-use but I might have missed some differences

client/src/components/Artist/ArtistSupport.tsx
fennifith: Ah, I'd planned on renaming this to `SubscriptionGrid` since it's used for artist subscriptions and not posts. The changes to the `<PostGrid>` component also make reuse less intuitive here, IMO.

simonv3 (approved):

PR #735, opened 2024-05-11
New variable to customize upload prompt description per image type

LouisLouiseKay:
I could have used directly "image type", but two limitations prompted me to create a new variable : 

- Incapacity to customize "a / an" based on the word type ("a background" "an avatar")
- Background is still defined as Banner in the backend, which is something we need to fix but is NOT a trivial rewrite because it's pretty much everywhere in the codebase and could have broken a lot of things. 

New behavior :

### DIFF
--- a/client/src/components/ManageArtist/UploadImage.tsx
+++ b/client/src/components/ManageArtist/UploadImage.tsx
@@ -9,6 +9,7 @@ import { FaFileUpload } from "react-icons/fa";
 
 export const Img = styled.img<{ rounded?: boolean }>`
   transition: .25s background-color, .25s filter;
+  aspect-ratio: 1/1;
 
   &:hover {
     filter: brightness(80%);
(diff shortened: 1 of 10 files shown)


client/src/components/ManageArtist/UploadImage.tsx
```diff
@@ -52,7 +52,8 @@ export const UploadPrompt: React.FC<{
   width?: string;
   height?: string;
   rounded?: boolean;
-}> = ({ width, height, rounded }) => {
+  imagetypedescription?: string;
```
simonv3, line 55: For consistency's sake and readability, can you make this camelcase? `imageTypeDescription`

simonv3 (commented):
looks good! For readability's sake could you make the variable camelcase?

LouisLouiseKay (commented):
I think it's ok now

Issue #520, opened 2024-01-17
Image uploaded as not square


simonv3:
![afbeelding](https://github.com/funmusicplace/mirlo/assets/485583/981dc753-1114-49d7-9865-d5cf3e643b73)

How do we prevent users from doing this? Do we need to? Not sure @LouisLouiseKay thoughts?

LouisLouiseKay:
Ah yeah, actually, it crops well to a square when the height is more than the width, so I'll have to give it some thoughts but it should be manageable, I'll have to work on that locally. 

The image uploader tool could probably crop on the way in the way it does it for the background, but I liked the idea of being able to have non-square images on the album page though, which would mean we only crop the thumbnails in the "releases / albums" sections.

samnabi:
<img width="1223" alt="Screen Shot 2024-03-16 at 12 30 26 PM" src="https://github.com/funmusicplace/mirlo/assets/1731536/982d338a-957a-4e11-b367-2e3dd15044f8">

Another example of a non-square album cover, for illustration. I do think it should be cropped to square for the releases/albums sections.

PR #736, opened 2024-05-11
Album image styling & cache-control headers

fennifith:
- Prevents the double image fetching in `<ImageWithPlaceholder>`
  * Since these requests were not cached, this was adding to the API ratelimit when clicking through release pages
- Ensures that image styling is square in release list & player (related: #520)
  * This uses `object-fit: contain;` rather than cropping the image, to ensure that the entire image is visible
  * Added lazy loading (which helps when images are out of view, although it *doesn't* help when they're already visible on the page - so it might be good to add a condition to this later)
  * Also added a fade-in transition when the image loads 
- Adds `Cache-Control` and `Etag` headers to responses from the `/images` endpoint
  * Also added naive support for `If-None-Match`, used with `stale-while-revalidate`

### DIFF
--- a/client/src/components/Artist/ArtistSquare.tsx
+++ b/client/src/components/Artist/ArtistSquare.tsx
@@ -24,6 +24,7 @@ const ArtistSquare: React.FC<{
           }
           alt={artist.name}
           size={300}
+          square
         />
 
         <TrackGroupLinks>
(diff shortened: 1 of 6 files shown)


simonv3:
Amazing thank you!

PR #737, opened 2024-05-11
Replace new-release route with tanstack query mutation

### LINKED ISSUE #732: Pressing 'back' when creating a new album creates another new album
The "add new album" button appears to link to a `/new-release` route that is kept in the browser history. This means that navigating backwards from the new album page revisits that route, rather than navigating back to the initial screen.

To reproduce:
- Go to an artist manage page, e.g. `/manage/artists/###/releases`
- Press the "+ Add new album" button
- Use the browser back button; this redirects back to the new album page
- Going back to the artist page shows that a bunch of draft releases have been created

Possible solutions:
- React router's `navigate()` has a [replace](https://reactrouter.com/en/main/hooks/use-navigate#optionsreplace) option that could be used to replace the history state with the created album URL, once known
- Album creation could occur within the new release button, avoiding the second redirect

fennifith:
- Instead of the `/new-release` route, adds a `<NewAlbumButton>` component that uses a mutation to create a new album
  - Eliminates the initial trackGroups fetch call by using a UUID as the temporary `urlSlug`
  - Fixes #732
- Likewise, uses a mutation for the album delete buttons
- Changes some button styling to avoid layout shifts when the `isLoading=` prop is changed
- Removes some now-unused behavior:
  - the `<NewReleaseRedirect>` component
  - `<NewAlbumForm>`
  - the `<Button>`'s `collapse=` prop and relevant styling, which does not appear to be in use


simonv3 (approved):

simonv3:
Looks great, thank you!

PR #756, opened 2024-05-25
Storybook configuration

fennifith:
- Configures [Storybook](https://storybook.js.org) with a few starting component stories: `<ArtistTrackGroup />`, `<PostCard />`, `<Button />`, and `<ButtonAnchor />`.
- Fixes a few type inconsistencies in TrackGroups (`cover.id` being a string, `cover.urls` being an array, & some properties not marked as optional)
- Migrates `<Releases />` to a `useQuery` call, and adds the `tag=` parameter to `queryTrackGroups`


simonv3 (approved):
lgtm! i wonder if we could set up netlify to auto-build the storybook so you could see it here and it could be used as a styleguide

fennifith:
Yep! `yarn client:storybook:build` will output to `client/storybook-static`, so I believe that can be set up by linking a second netlify site to this repo and using that as the build command/dir.

PR #832, opened 2024-09-02
Show color invalid error

### LINKED ISSUE #779: Catch invalid colors
> if you input an invalid color here (by removing the hashtag, for instance), the corresponding elements become invisible

We should use this regex on validation

```
"^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$"
```

peterklingelhofer:
Closes #779

### DIFF
--- a/client/src/components/ManageArtist/ColorInput.tsx
+++ b/client/src/components/ManageArtist/ColorInput.tsx
@@ -15,7 +15,7 @@ const generateColor = (name: string) => {
 };
 
 const isValidColor = (val: string) => {
-  const matcher = val.match(/^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/);
+  const matcher = val.match(/^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3}|[A-Fa-f0-9]{8}|[A-Fa-f0-9]{4})$/);
   return matcher;
 };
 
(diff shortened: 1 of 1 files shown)


client/src/components/ManageArtist/ColorInput.tsx
```diff
@@ -15,7 +15,7 @@ const generateColor = (name: string) => {
 };
 
 const isValidColor = (val: string) => {
-  const matcher = val.match(/^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/);
+  const matcher = val.match(/^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3}|[A-Fa-f0-9]{8}|[A-Fa-f0-9]{4})$/);
```
peterklingelhofer, line 18:
```suggestion
  const matcher = val.match(/^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/);
```

client/src/components/ManageArtist/ColorInput.tsx
```diff
@@ -50,7 +50,7 @@ export const ColorInput: React.FC<{ name: string; title: string }> = ({
   React.useEffect(() => {
     clearErrors(name);
     if (!isValidColor(color)) {
-      setError(name, { message: "Not a valid color" });
+      setError(name, { message: "Not a valid color code. Please use #RRGGBB, #RGB, #RRGGBBAA, or #RGBA format." });
```
peterklingelhofer, line 53:
```suggestion
      setError(name, { message: "Not a valid color code. Please use #RRGGBB, #RGB format." });
```

simonv3:
My thought here was that this would show that a color is not valid as a user is typing it? Maybe that's already the case though

PR #877, opened 2024-10-10
uh.....

humancuration:
tests from jest to sinon
files related to 257, 509, 422, 754

no idea if any of these will be helpful or not but i guess this is a pull request, thanks

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -29,64 +29,79 @@ function App() {
 
   return (
     <>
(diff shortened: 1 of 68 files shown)


simonv3:
Thanks for this! As is this will definitely break our app. I think it might be helpful here to pick very specific small changes that you'd like to see and put in individual PRs for each of those. That way we can test how they work out without impacting other parts of the system.

PR #887, opened 2024-10-20
Include coverart in downloads

### LINKED ISSUE #645: Downloads don't include cover art
I expected the download zips to include cover art (often called cover.jpg or folder.jpg).

ksa618:
Includes the 1500 pixel cover image as "cover.jpg" in the zip file when downloading.

Closes #645

### DIFF
--- a/src/jobs/generate-album.ts
+++ b/src/jobs/generate-album.ts
@@ -6,6 +6,7 @@ import { logger } from "./queue-worker";
 import {
   createBucketIfNotExists,
   finalAudioBucket,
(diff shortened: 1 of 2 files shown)


simonv3 (approved):
Amazing, thank you!

Not for this PR, but it'd be cool if we could write some tests for these tasks to see if they stay working.

ksa618:
I'll look into the tests.

One thing about the coverart. The zip-files need to be regenerated to include the cover files, so the previously generated zip-files stored in minio in production will have to be deleted.

PR #892, opened 2024-10-23
Add artist name to zip

### LINKED ISSUE #889: Include artist name in album download name
> This might just be me and it definitely is very petty to point it out (sorry!) but I hate it that when I download an album the zip file doesn't include the artist name.

Should be an easy fix!

ksa618:
Closes #889

The inclusions in `basicTrackGroupInclude` keeps growing. I don't know if that's a problem.

### DIFF
--- a/src/routers/v1/trackGroups/{id}/download.ts
+++ b/src/routers/v1/trackGroups/{id}/download.ts
@@ -99,7 +99,7 @@ export default function () {
       }
 
       try {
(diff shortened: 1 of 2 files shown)


simonv3 (approved):

simonv3:
Yeah it's probably not great long term if we're using the trackGroupInclude in `findMany`for trackGroups, but that's pretty limited at the moment. it probably makes sense to construct that trackGroupInclude with some variables, or just be explicit every time we call it. Maybe verbosity is actually better here.

PR #893, opened 2024-10-23
Fix link display

ksa618:
I noticed some weirdness with the displayed links on the artist page: 
image

I cleaned this up, to use the link type that's stored in the database as link title. A further improvement would be to add a possibility to choose "Other" and type the link name in the the "Edit links" dialog. But that's an exercise left for another time.


client/src/components/common/LinkIconDisplay.test.tsx
```diff
@@ -37,3 +37,35 @@ test("linkUrlHref doesn't affect emails that already have a mailto prefix", () =
   const result = linkUrlHref("mailto:test@example.com");
   expect(result).toBe("mailto:test@example.com");
 });
+
```
simonv3, line 40: 😍

simonv3 (approved):

PR #903, opened 2024-11-02
Add API tests for artist/id/subscribe using stripe-mock

ksa618:
I had to change from `res.status(...)` to `ApiError` for errors. The tests would just hang, waiting for the call to return before I did the change.

### DIFF
--- a/src/utils/stripe.ts
+++ b/src/utils/stripe.ts
@@ -18,12 +18,16 @@ import {
 } from "./handleFinishedTransactions";
 import countryCodesCurrencies from "./country-codes-currencies";
 
-const { STRIPE_KEY, API_DOMAIN } = process.env;
+// STRIPE_HOST, STRIPE_PORT, and STRIPE_PROTOCOL should not be set when not running tests against stripe-mock
+const { STRIPE_KEY, API_DOMAIN, STRIPE_HOST, STRIPE_PORT, STRIPE_PROTOCOL } = process.env;
 
 export const OPTION_JOINER = ";;";
 
 const stripe = new Stripe(STRIPE_KEY ?? "", {
   apiVersion: "2022-11-15",
+  host: STRIPE_HOST,
+  port: STRIPE_PORT,
+  protocol: STRIPE_PROTOCOL === "http" ? "http" : "https"
 });
 
 const calculatePlatformPercent = async (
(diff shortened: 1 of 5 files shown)


src/utils/stripe.ts
```diff
@@ -18,12 +18,16 @@ import {
 } from "./handleFinishedTransactions";
 import countryCodesCurrencies from "./country-codes-currencies";
 
-const { STRIPE_KEY, API_DOMAIN } = process.env;
+// STRIPE_HOST, STRIPE_PORT, and STRIPE_PROTOCOL should not be set when not running tests against stripe-mock
+const { STRIPE_KEY, API_DOMAIN, STRIPE_HOST, STRIPE_PORT, STRIPE_PROTOCOL } = process.env;
```
simonv3, line 22: should these be included in .env.example as comments? or can we do an if statement that's something if ENV === 'test' or CI = true?

simonv3 (commented):
lgtm but do have that one question

src/utils/stripe.ts
ksa618: I don't think they should be in .env.example. The stripe-mock is only for automatic tests, so it wouldn't make sense to set them when running the api containers for manual testing. It's better to use stripe test mode there. Having an if statement is probably a good idea. I'll change it to only use those env variables in test mode.

ksa618: I tried it and it works, but I have to set `NODE_ENV: test` for `api-test` in the docker-compose.test.yml file (it's currently "development"). The tests pass when I do this, but I see there are a few checks for `NODE_ENV === "development"` in the code. Is it a problem that it's "test" when the tests run?

simonv3: I'm looking at the code and I think if the tests pass I think it's okay. The node_env === 'dev stuff looks like it's mainly to make some things "more permissive" like to not have rate limiting, for cors (not an issue in the back-end tests). So I think it's okay?

simonv3 (approved):

src/utils/stripe.ts
simonv3: does the stripe image load these into the env environment? I guess I'm not clear on where the environment variables are coming from

ksa618: I've set the in docker-compose.test.yml.

simonv3 (approved):

src/utils/stripe.ts
simonv3: Amazing! feel free to merge it when you're happy with it!

PR #948, opened 2024-12-19
WIP: switch to using backblaze for images

simonv3:
This is a punt that switches to using backblaze for images

### DIFF
--- a/src/config/sharp.ts
+++ b/src/config/sharp.ts
@@ -69,6 +69,7 @@ export default {
     avatar: {
       webp: {
         variants: [
+          { width: 3000, height: 3000 },
           { width: 1500, height: 1500 },
           { width: 1200, height: 1200 },
           { width: 960, height: 960 },
--- a/src/config/sharp.ts
+++ b/src/config/sharp.ts
@@ -11,30 +11,30 @@ export default {
         smartSubsample: true,
       },
     },
-    jpeg: {
-      ext: ".jpg",
-      outputOptions: {
-        chromaSubsampling: "4:4:4",
-        optimiseCoding: true,
-        quantisationTable: 3,
-        progressive: false,
-        optimiseScans: false, // required mozjpeg support
-        trellisQuantisation: true, // required mozjpeg support
-        overshootDeringing: true, // required mozjpeg support
-      },
-    },
-    png: {
-      ext: ".png",
-      outputOptions: {
-        chromaSubsampling: "4:4:4",
-        optimiseCoding: true,
-        quantisationTable: 3,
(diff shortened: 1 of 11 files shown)


src/config/sharp.ts
```diff
@@ -11,30 +11,30 @@ export default {
...
-        chromaSubsampling: "4:4:4",
-        optimiseCoding: true,
-        quantisationTable: 3,
-        progressive: false,
-        optimiseScans: false, // required mozjpeg support
-        trellisQuantisation: true, // required mozjpeg support
-        overshootDeringing: true, // required mozjpeg support
-      },
-    },
+    // jpeg: {
```
simonv3, line 14: So I'm getting rid of these because they're not getting used in the UI, and I can't quite imagine why they'd get used, and we can always generate them in the future?

src/jobs/optimize-image.ts
```diff
@@ -40,42 +42,53 @@ const optimizeImage = async (job: Job) => {
...
+            },
+            j
+          ) => {
+            // This is a pretty hacky way of sleeping and probably something we shoudl
+            // not do?
+            // But basically the reason we have to do this is because backblaze
+            // seems to not like it when we bombard it with files all in one go,
+            // so when we do Promise.all at the end of this we're sending all the
+            // files and it just doesn't process anything but the first one.
+            await sleep((i + 1) * (j + 1) * 1500);
```
simonv3, line 91: Having to sleep in these promises because if I don't then backblaze seems to get overwhelmed and returns an error 408

src/static.ts
```diff
@@ -1,17 +1,38 @@
...
-  let stat;
+  let backblazeStat;
+  let minioStat;
+  logger.info(`static: ${req.params.bucket}/${req.params.filename}`);
+
   // FIXME: someone who's better at devops than me will have to figure out
   // how we can serve these directly from minio
   try {
-    stat = await minioClient.statObject(req.params.bucket, req.params.filename);
+    backblazeStat = await backblazeClient.send(
```
simonv3, line 23: While this works it feels pretty slow. I'm wondering if making the image public on backblaze and referencing it directly would be better for us and our bandwidth. 

I guess in that case we need to change the `generateFullStaticImageUrl` url to reference backblaze, in which case it might make sense to store on our image object where it is stored, which would be a boolean I guess?

simonv3: After some deep debugging it looks like docker was having issues resolving addresses outside of its own network, so when I changed the docker DNS to just be the google IP (and not include 10.0.0.2 as it was on my computer). This seemed to work much faster.

PR #963, opened 2025-01-11
Adds the subscription and purchase pages

### LINKED ISSUE #257: Add subscription and purchase success pages
Right now the `checkout` route directs the front-end to the same page as the album or subscription page, but we'd be better off creating a specific page for purchase and subscription confirmation, just so that it also works for non-logged in users. 

1. [ ] Create a new route in the front-end for each of `purchase` and `subscribe` processes (maybe it can handle both?) 
2. [ ] Update the `checkout` route to redirect the new place

lenikadali:
Fixes #257

### DIFF
--- a/client/src/components/Tip/CheckoutComplete.tsx
+++ b/client/src/components/Tip/CheckoutComplete.tsx
@@ -1,0 +1,86 @@
+import { css } from "@emotion/css";
+
+import { useParams } from "react-router-dom";
+import Box from "../common/Box";
+import { useTranslation } from "react-i18next";
+import FullPageLoadingSpinner from "components/common/FullPageLoadingSpinner";
+import ImageWithPlaceholder from "components/common/ImageWithPlaceholder";
+import { WidthWrapper } from "components/common/WidthContainer";
+
+import { useQuery } from "@tanstack/react-query";
+import { queryArtist, queryMerch } from "queries";
+import Confetti from "./Confetti";
+
+function TrackGroup() {
+  const { t } = useTranslation("translation", {
+    keyPrefix: "merchDetails",
+  });
+
+  const { artistId, merchId } = useParams();
+  const { data: artist, isLoading: isLoadingArtist } = useQuery(
+    queryArtist({ artistSlug: artistId ?? "" })
+  );
+  const { data: merch, isLoading: isLoadingMerch } = useQuery(
+    queryMerch({ merchId: merchId ?? "" })
+  );
+
+  if (!artist && !isLoadingArtist) {
+    return <Box>{t("doesNotExist")}</Box>;
+  } else if (!artist) {
+    return <FullPageLoadingSpinner />;
+  }
+
+  if (!merch && !isLoadingMerch) {
+    return <Box>{t("doesNotExist")}</Box>;
+  } else if (!merch) {
+    return <FullPageLoadingSpinner />;
+  }
+
+  return (
+    <WidthWrapper
+      variant="medium"
+      className={css`
+        margin-top: 4rem !important;
+        align-items: center;
+
+        h1 {
+          margin-bottom: 1rem;
+        }
+
+        svg {
+          max-width: 200px;
+          margin: 0 auto;
+          display: block;
+        }
+      `}
+    >
+      <h1>Purchase complete!</h1>
+      <div
+        className={css`
+          display: flex;
+
+          div:first-child {
+            flex-grow: 1;
+          }
+        `}
+      >
+        <ImageWithPlaceholder
+          src={merch.images?.[0]?.sizes?.[120]}
+          alt={merch.title}
+          size={120}
+        />
+        <div
+          className={css`
+            margin-left: 1rem;
+          `}
+        >
+          {" "}
+          You've bought {merch.title}! The artist has been notified.
+        </div>
+      </div>
+      <Confetti />
+    </WidthWrapper>
+  );
+}
+
+export default TrackGroup;
--- a/client/src/routes.tsx
+++ b/client/src/routes.tsx
@@ -651,6 +651,27 @@ const routes: RouteObject[] = [
           },
         ],
       },
+      {
+        path: "artistSlug",
+        children: [
+          {
+            path: "tip",
+            children: [
+              {
+                path: ":tipId/checkout-complete",
+                async lazy() {
+                  const { default: Component } = await import(
+                    "components/Tip/CheckoutComplete"
+                  );
+                  return {
+                    Component: () => <Component />,
+                  };
+                },
+              },
+            ],
+          },
+        ],
+      },
     ],
   },
 ];
--- a/client/src/components/Tip/Confetti.tsx
+++ b/client/src/components/Tip/Confetti.tsx
@@ -1,0 +1,403 @@
+import { css } from "@emotion/css";
+
+const Confetti: React.FC = () => {
+  return (
+    <svg
+      version="1.1"
+      id="Layer_1"
+      xmlns="http://www.w3.org/2000/svg"
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+      x="0px"
+      y="0px"
+      viewBox="0 0 400 400"
+      // @ts-ignore
+      style={{ enableBackground: "new 0 0 800 800" }}
+      className={css`
+        .conf0 {
+          fill: #fc6394;
+        }
+        .conf1 {
+          fill: #ef3c8a;
+        }
+        .conf2 {
+          fill: #5adaea;
+        }
+        .conf3 {
+          fill: #974cbe;
+        }
+        .conf4 {
+          fill: #3cbecd;
+        }
+        .conf5 {
+          fill: #813bbe;
+        }
+        .conf6 {
+          fill: #f9b732;
+        }
+        .conf7 {
+          display: none;
+          fill: none;
+          stroke: #000000;
+          stroke-miterlimit: 10;
+        }
+        .conf8 {
+          fill: none;
+          stroke: #f9b732;
+          stroke-width: 9;
+          stroke-linecap: round;
+          stroke-miterlimit: 10;
+        }
+
+        .confetti-cone {
+          transform-origin: 200px 50px;
+          animation: confetti-cone1 1.2s ease infinite;
+        }
+        @keyframes confetti-cone1 {
+          0% {
+            transform: translate(40px, 95px) rotate(45deg) scale(1, 1);
+          }
+          15% {
+            transform: translate(10px, 145px) rotate(45deg) scale(1.1, 0.85);
+          }
+          100% {
+            transform: translate(40px, 105px) rotate(45deg) scale(1, 1);
+          }
+        }
+
+        #yellow-strip {
+          fill: none;
+          stroke: #f9b732;
+          stroke-width: 9;
+          stroke-linecap: round;
+          stroke-miterlimit: 10;
+          animation: confdash 1.2s ease infinite;
+        }
+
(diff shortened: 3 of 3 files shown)


client/src/components/Tip/CheckoutComplete.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { useParams } from "react-router-dom";
+import Box from "../common/Box";
+import { useTranslation } from "react-i18next";
+import FullPageLoadingSpinner from "components/common/FullPageLoadingSpinner";
+import ImageWithPlaceholder from "components/common/ImageWithPlaceholder";
+import { WidthWrapper } from "components/common/WidthContainer";
+
+import { useQuery } from "@tanstack/react-query";
+import { queryArtist, queryMerch } from "queries";
+import Confetti from "./Confetti";
```
lenikadali, line 12: I am not sure of how `tip` relates to the artist. Will take a closer look but the queries may need to be updated in order to get the tip information and use it in the component.

Also I think since we're using the `Confetti` component here as well, it is probably a sign that we should move into `common` so that other components can use it in future.

client/src/components/Tip/CheckoutComplete.tsx
```diff
@@ -0,0 +1,86 @@
...
+    return <Box>{t("doesNotExist")}</Box>;
+  } else if (!artist) {
+    return <FullPageLoadingSpinner />;
+  }
+
+  if (!merch && !isLoadingMerch) {
+    return <Box>{t("doesNotExist")}</Box>;
+  } else if (!merch) {
+    return <FullPageLoadingSpinner />;
+  }
```
lenikadali, line 37: Related to the comment above

client/src/routes.tsx
```diff
@@ -651,6 +651,27 @@ const routes: RouteObject[] = [
           },
         ],
       },
+      {
+        path: "artistSlug",
```
simonv3, line 655: You could nest this under `:artistId`

client/src/components/Tip/CheckoutComplete.tsx
```diff
@@ -0,0 +1,80 @@
...
+    keyPrefix: "tipArtist",
+  });
+
+  const { artistId, artistSlug, tipId } = useParams();
+  const { data: artist, isLoading: isLoadingArtist } = useQuery(
+    queryArtist({ artistSlug: artistId ?? "" })
+  );
+  const { data: tip, isLoading: isLoadingTip } = useQuery(
+    queryTip({ artistId: artistId, artistSlug: artistSlug, tipId: tipId })
+  );
```
lenikadali, line 25: Not sure if this is doing what is intended. But this is a first pass at demonstrating what the query behaviour should look like.

client/src/components/Tip/CheckoutComplete.tsx
```diff
@@ -0,0 +1,80 @@
...
+        className={css`
+          display: flex;
+
+          div:first-child {
+            flex-grow: 1;
+          }
+        `}
+      >
+        <ImageWithPlaceholder
+          src={artist.avatar?.[0]?.sizes?.[120]}
```
lenikadali, line 62: In theory this should work but in VSCode, I get the following message when I hover over the `{artist.avatar?.[0]` portion which has a red line underneath:

```
Element implicitly has an 'any' type because expression of type '0' can't be used to index type '{ url: string; sizes?: { [key: string]: string; } | undefined; updatedAt: string; }'.
  Property '0' does not exist on type '{ url: string; sizes?: { [key: string]: string; } | undefined; updatedAt: string; }'.
```

client/src/queries/tip.ts
```diff
@@ -0,0 +1,33 @@
...
+  ["fetchTip", { artistId: string; artistSlug: string; tipId: string }]
+> = ({ queryKey: [_, { artistId, artistSlug, tipId }], signal }) => {
+  return api
+    .get<{
+      result: UserArtistTip;
+    }>(`/v1/artists/${artistId}/${artistSlug}/${tipId}`, {
+      signal,
+    })
+    .then((r) => r.result);
+};
```
lenikadali, line 15: I'm making assumptions here that I am not sure about. Probably needs to me to take a look at `UserArtistTip` and see if there's a shorter way to do this (I feel using `artistId` is a code smell)

lenikadali (commented):
Updated the PR with a first pass at writing the query code and addressing the comments from the first round of review :raised_hands:

client/src/components/Tip/CheckoutComplete.tsx
simonv3: There'd only be one avatar picture, so I think it should be `avatar?.sizes`

client/src/queries/tip.ts
simonv3, line 15: you don't need both `artistUrl` and `artistId`, the api handles both (because `artistSlug` is unique). 

```suggestion
const fetchTip: QueryFunction<
  UserArtistTip,
  ["fetchTip", { artistId: number; tipId: string }]
> = ({ queryKey: [_, { artistId, tipId }], signal }) => {
  return api
    .get<{
      result: UserArtistTip;
    }>(`/v1/artists/${artistId}/${tipId}`, {
      signal,
    })
    .then((r) => r.result);
};
```

However, I think you'll need to add the endpoint for this in the API.

client/src/queries/tip.ts
```diff
@@ -0,0 +1,33 @@
...
+      "fetchTip",
+      {
+        artistId: opts.artistId,
+        artistSlug: opts.artistSlug,
+        tipId: opts.tipId,
+      },
+    ],
+    queryFn: fetchTip,
+  });
+}
```
simonv3, line 33: Using the idea that we'll implement an endpoint to fetch the id directly (as outlined above) `manage/tips/:tipId`

```suggestion
export function queryTip(opts: {
  tipId: string;
}) {
  return queryOptions({
    queryKey: [
      "fetchTip",
      {
        tipId: opts.tipId,
      },
    ],
    queryFn: fetchTip,
  });
}
```

client/src/components/Tip/CheckoutComplete.tsx
lenikadali: Updated version in [9e4d9a1](https://github.com/funmusicplace/mirlo/pull/963/commits/9e4d9a15e6056562c5de0adaf7cbcc72ff08a7cb)

client/src/queries/tip.ts
lenikadali: Superseded by [6ac18c4](https://github.com/funmusicplace/mirlo/pull/963/commits/6ac18c4c777ec8d847ee34ffe64e9879560cefb2)

lenikadali: Resolved in [6ac18c4](https://github.com/funmusicplace/mirlo/pull/963/commits/6ac18c4c777ec8d847ee34ffe64e9879560cefb2)

client/src/queries/tip.ts
lenikadali: Resolved in [6ac18c4](https://github.com/funmusicplace/mirlo/pull/963/commits/6ac18c4c777ec8d847ee34ffe64e9879560cefb2)

client/src/components/Tip/CheckoutComplete.tsx
```diff
@@ -0,0 +1,80 @@
...
+    keyPrefix: "tipArtist",
+  });
+
+  const { artistId, tipId } = useParams();
+  const { data: artist, isLoading: isLoadingArtist } = useQuery(
+    queryArtist({ artistSlug: artistId ?? "" })
+  );
+  const { data: tip, isLoading: isLoadingTip } = useQuery(
+    queryTip({ tipId: tipId })
+  );
```
lenikadali, line 25: For the lines:
```javascript
  const { data: artist, isLoading: isLoadingArtist } = useQuery(
    queryArtist({ artistSlug: artistId ?? "" })
  );
```
should `queryArtist({ artistSlug: artistId ?? "" })` be changed to `queryArtist({ artistId: artistId ?? "" })`?

The lines

```javascript
const { data: tip, isLoading: isLoadingTip } = useQuery(
    queryTip({ tipId: tipId })
  );
```
have the  `const { data: tip, isLoading: isLoadingTip }` greyed out as they are unused. I wonder if we should modify it to

```javascript
const { data: tip } = useQuery(
    queryTip({ tipId: tipId })
  );
```
instead? Also the `queryTip({ tipId: tipId })` is failing on a type error:
```
Type 'string | undefined' is not assignable to type 'string'.
```
not sure how to modify the caller without painting myself into a corner. Thoughts?

simonv3: > should queryArtist({ artistSlug: artistId ?? "" }) be changed to queryArtist({ artistId: artistId ?? "" })?

Using `artistSlug` should be fine. Either way it'll be a string, and the query uses `artistSlug`, and the API can handle either one.

> have the const { data: tip, isLoading: isLoadingTip } greyed out as they are unused. I wonder if we should modify it to [etc]

That's your call. If you don't want to show a loading indicator while this loads, then you don't need to.

> Also the queryTip({ tipId: tipId }) is failing on a type error: `Type 'string | undefined' is not assignable to type 'string'.`

The two options here are to let the queryTip function accept an undefined value, or to nest the component in a sub component for which you guarantee that the tip exists. Psuedocode:

```
const { tipId } = useParams();

if (!tipId) { return null }

return <TipComponent id={tipId} />
```

and then the tipComponent: 

```
const TipComponent: React.FC<{tipId: string}> = { tipId } => {

// other process
}
```

lenikadali: Opted to just do the `null` check before querying for the tip as that seemed to remove the type issue and refactoring into a separate component looked more complex.

client/src/components/Tip/CheckoutComplete.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+        svg {
+          max-width: 200px;
+          margin: 0 auto;
+          display: block;
+        }
+      `}
+    >
+      <h1>You have successfully tipped this artist!</h1>
+      <div
```
simonv3, line 55: Likewise I wonder if this can be turned into a re-used component where you pass the `artist.name` and `imageSrc`

client/src/routes.tsx
```diff
@@ -194,6 +194,22 @@ const routes: RouteObject[] = [
               return { Component };
             },
           },
+          {
+            path: "tips",
```
simonv3, line 198: Great! I think also the checkout process on the api in `src/routers/v1/checkout` will need to be updated to point to the right page. You can see it pointing to the right page on line 54 for merch, and it will need to be updated for line 63 too.

simonv3 (commented):
Sorry for the slow reply, it's starting to look good! I've got one clean up comment and one comment that would tie it in so that people see the page when they're done with checkout!

client/src/routes.tsx
lenikadali: Hopefully I got it right here [cf05a98](https://github.com/funmusicplace/mirlo/pull/963/commits/cf05a98e6d848d750ea85f927e73d22b18234b3a)

client/src/components/Tip/CheckoutComplete.tsx
lenikadali: I am leaning towards making a follow-up PR for this and [the comment above](https://github.com/funmusicplace/mirlo/pull/963/files/0574a76eeb885a8731d6ff550223162cad667491#r2039730750)
Mostly because I feel this would be too much change in one PR and it would be clearer how to rework both components into one that can be re-used across both Tips and Merchandise.

PR #1053, opened 2025-03-11
Replace download button with a link, to avoid popup blocker problems

### LINKED ISSUE #644: Firefox popup blocker blocks album download
When I tried to download the album I bought, I got this "We're generating the album! Hold on for a minute..." message:

Screenshot from 2024-04-04 23-10-13

It took several minutes, actually. Quite slow. (I hope optimizing this is on your radar.) So I switched tabs and was doing something else. When I tabbed back, that message had simply disappeared, and there was no indication of why it had disappeared. After some confusion I decided to click Download a second time and it worked.

To try to repro this and give you a more useful bug report, I tried again with a different file format and this time kept watching the tab. It turns out Firefox's popup blocker is blocking the delayed download:

Screenshot from 2024-04-04 23-11-57

Firefox's behavior makes sense to me, for what it's worth. It would be confusing to have a file download suddenly interrupt me several minutes later when doing something else.

I suggest redesigning this UI so there's only a button to click _after_ the files are ready. Start converting the files as soon as a file format is selected, and show a spinner (better yet, progress bar with ETA) + disabled button until conversion is done.

ksa618:
Closes #644, hopefully

### DIFF
--- a/client/src/components/common/DownloadAlbumButton.tsx
+++ b/client/src/components/common/DownloadAlbumButton.tsx
@@ -1,4 +1,3 @@
-import Button from "components/common/Button";
 import React from "react";
 import { css } from "@emotion/css";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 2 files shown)


simonv3 (approved):
oh woops sorry for the delay on the review

PR #1142, opened 2025-05-10
Update PostHeader.tsx

tmray:
On posts with no images, on mobile and desktop, the height for the h1 container divs was reversed. 

It should be height: 20vh on mobile and height: 30vh on desktop.

I've added those changes here.


simonv3:
lgtm, thank you!!

PR #1178, opened 2025-05-26
Buying something using a screen reader and/or keyboard

### LINKED ISSUE #1157: Issue pressing buy button with a screen reader
> Apparently with her screenreader (she’s blind) it’s not possible to actually click the “Buy” button.
> (cf: [SoNoMu](https://sonomu.club/@DasNest@masto.ai/114499588475006270) ; German only – sorry)
> Don’t know if this is the right place to mention it, but I thought I’d mention it somewhere before it gets lost.
> Keep up the really great work, it’s nice to see this growing as a platform!

timglorioso:
Fix #1157 

Todo:

- [x] Add focus trap to Modal
- [x] Use title for Modal label (`aria-labelledby`)
- [x] Allow passing ID of children describing the Modal via `ariaDescribedBy`
- [x] Close Modal on escape
- [x] Return focus to Modal initiator
- [x] Test flow with screen reader
- [x] Enable buy button in `BuyTrackGroup` dynamically (see [c458e7a](https://github.com/funmusicplace/mirlo/pull/1178/commits/c458e7aaeac752dbc87cac365a5e5d305e076b66))
- [x] ~Cypress?~
- [x] Add `label` to price field in `BuyTrackGroup`
- [x] Prevent scrolling outside of Modal
- [x] Test logged out


simonv3:
@timglorioso I think for Cypress to work properly we still have to do some devops fine-tuning. Right now it just browses the internet haha. I don't know if you want to bite that off for this PR so feel free to not write a test for it.

client/src/components/common/Modal.tsx
```diff
@@ -163,54 +177,69 @@ export const Modal: React.FC<{
...
+				role="dialog"
+				aria-describedby={ariaDescribedBy || undefined}
+				aria-labelledby={dialogLabelFromTitle(title)}
+				aria-modal="true"
+				onKeyDown={(event) => {
+					event.key === "Escape" && onCloseWrapper(event)
+				}}
+			>
+				<Background onClick={onCloseWrapper} />
+				<div className={wrapper} data-cy="modal">
```
timglorioso, line 196: Thanks @simonv3 for the info about Cypress. I saw this `data-cy` and wasn't sure if it might be moved to the `div role=dialog` above it. I guess if the tests don't do much now, there's not much risk either way.

timglorioso:
I'm considering this PR ready to merge. I tested the Buy flow with VoiceOver. Some more notes below.

My original fix for the Buy button state (https://github.com/funmusicplace/mirlo/pull/1178/commits/1637c06cc685fbad4737aef4eb4298bdcb8052cb) didn't work for logged out users. The button ended up enabled without an email entered.

I found this tidbit from the [React Hook Form docs on `formState`](https://react-hook-form.com/docs/useform/formstate) to fix it by destructuring:

```
// ❌ formState.isValid is accessed conditionally,

// so the Proxy does not subscribe to changes of that state
return <button disabled={!formState.isDirty || !formState.isValid} />;

// ✅ read all formState values to subscribe to changes
const { isDirty, isValid } = formState;
return <button disabled={!isDirty || !isValid} />;
```

On a separate note, [we should keep form controls enabled](https://adrianroselli.com/2024/02/dont-disable-form-controls.html) and provide feedback when changes are needed to proceed. In this case, the ideal is to validate the form on Buy button click and then display any errors. I actually wrote this code, but it touches more parts of `BuyTrackGroup` internals and experience and is worth a separate review/PR.

PR #1196, opened 2025-06-02
Attach trackGroup -> cover field to Each Track in db query

rann143:
- Tracks returned from the get all tracks db query will have their trackGroups' cover field attached


src/routers/v1/tracks/index.ts
```diff
@@ -36,6 +38,7 @@ export default function () {
           trackGroup: {
             include: {
               artist: true,
+              cover: true, //added this line
```
simonv3: You can remove the comment `//added this line` :)

simonv3 (approved):

PR #1230, opened 2025-06-09
Create topSold Endpoints for Tracks & TrackGroups

rann143:
Creates topSold Endpoints listed issue #1176

PR #1259, opened 2025-06-18
Create mostPlayed Endpoints for Tracks and TrackGroups

### LINKED ISSUE #1176: Add "best selling" and "top listened" endpoints
For both trackgroups and tracks. 

I think it makes sense to contain these in their own endpoints. 

`v1/tracks/mostPlayed`

`v1/trackGroups/mostPlayed`

`v1/tracks/topSold`

`v1/trackGroups/topSold`

rann143:
With PR #1230, this PR closes #1176 

- Adds mostPlayed endpoints for tracks and trackGroups
- Tests created and passed for both endpoints


simonv3 (commented):
This is really great but I'm pretty worried that this is going to be a pretty big query as the size of the database grows. I'm wondering if there's a more efficient [raw query}(https://www.prisma.io/docs/orm/prisma-client/using-raw-sql/raw-queries) for this?

simonv3 (approved):
Let's see what happens!! If we see it seriously impacting our performance we can re-evaluate.

PR #1268, opened 2025-06-22
Create Verbal communication guidelines (words we choose and why).md

Cwpute:
Added communication guidelines doc. hurray !
Not sure if this is the right place to put this doc. The name might be subject to change if there is a naming convention i'm not aware of (although i like the "words we choose and why" bit).


simonv3:
Looks great!!

PR #1394, opened 2025-08-03
added artistName variable to multiple strings +minor wording

Cwpute:
added a point at the end of a sentence.
some could say it was... pointless (but hopefully not this commit)

EDIT: now included, lots of artistName variable addition in strings that would have needed it.


Cwpute:
idk why but it seems to have added my latest submissions to this old pull request :shrug: hopefully this is alright !
everything done by hand through Github's file editor... maybe if i continue doing these sort of things i should use more professional tools.

To note : Came upon a case where `artist.chooseSupportLevel` already had the variable declared but with `artist?.name` instead of `artist.name`... i suppose this is a wildcard symbol to support multiple use cases ? but with my level in coding i can only guess. I left it as is and didn't add anything, but it may have to be corrected... or may my own editions should be modified themselves !... let me know in any case.

client/src/components/TrackGroup/FreeDownload.tsx
```diff
@@ -42,7 +42,7 @@ const FreeDownload: React.FC<{
           </strong>
           <form style={{ marginTop: "1rem" }}>
             <p style={{ marginBottom: "1rem" }}>
-              {t(user ? "justDownloadLoggedIn" : "justDownloadNoUser")}
+              {t(user ? "justDownloadLoggedIn", {artistName: artist.name} : "justDownloadNoUser", {artistName: artist.name})}
```
simonv3: I think here you have to reference the trackGroup.artist
```suggestion
              {t(user ? "justDownloadLoggedIn", {artistName: trackGroup.artist.name} : "justDownloadNoUser", {artistName: trackGroup.artist.name})}
```

client/src/components/common/FollowArtistNotLoggedInForm.tsx
```diff
@@ -28,7 +28,7 @@ const FollowArtistNotLoggedInForm: React.FC<{ artistId: number }> = ({
 
   return (
     <form>
-      <p>{t("toFollowThisArtist")}</p>
+      <p>{t("toFollowThisArtist", {artistName: artist.name})}</p>
```
simonv3, line 31: The artist isn't defined in this component, but actually, I don't think this component is used anywhere, so we should just ignore it (I can delete it myself or try and find _where_ it shoudl be removed.

simonv3 (changes requested):
Largely looks good to me, but there's two bits where a change is needed.

client/src/components/common/FollowArtistNotLoggedInForm.tsx
Cwpute: i don't know how artists are defined in any component, so please do whatever feels better to you.
and with that the commit will be ready to be merged right ?
ah no, there are conflicts but i don't know how to resolve those either. i think these might be files that changed either name, place or were removed entirely probably...

simonv3:
I merged this in #1409 to fix the merge conflicts

PR #1405, opened 2025-08-11
Support for unicode characters in titles of tracks and albums

shehackedyou:
I am not able to test this, because I don't really want to setup a server but this provides 3 fixes to the audio processing pipeline that should enable unicode support

If you have questions about my recommended changes, let me know


simonv3 (approved):
Pulled this down and afaik looks good! I'll merge and @shehackedyou if you want to try it once it's deployed (usually around 20 minutes after merge) let us know!

PR #1415, opened 2025-08-14
Email Album Link for Purchase

rann143:
# Changes:
- Create new API endpoint that sends an email with a link to the selected album's page where the user can make a purchase
- Create new email template for delivering this information

# Reason:

For the mobile app. Due to Apple's guidelines, we're not allowed to encourage an alternative method of purchase other than in-app purchases due to [Apple's requirements](https://developer.apple.com/app-store/review/guidelines/#other-purchase-methods) - so we can't link to our website for people to buy whichever track/album they were listening to.

We'd need to have a "Email Purchase Info" button instead, which opens a small form like below for the user's email. This would fall within Apple's guidelines.


emails/album-purchase-link/html.pug
```diff
@@ -0,0 +1,31 @@
+doctype html
```
simonv3: You can look at `emails/announce-post-published` to see us use the layout.pug template. A bit better for code re-use and lets us apply styles to all emails.

Not all e-mails have made it to have that layout yet. Have been migrating it as I come across them.

src/routers/v1/trackGroups/{id}/emailPurchaseLink.ts
```diff
@@ -0,0 +1,76 @@
+import { NextFunction, Request, Response } from "express";
+import { userLoggedInWithoutRedirect } from "../../../../auth/passport";
+import prisma from "@mirlo/prisma";
+
+import sendMail from "../../../../jobs/send-mail";
+import { Job } from "bullmq";
+
+export default function () {
+  const operations = {
+    POST: [userLoggedInWithoutRedirect, POST],
```
simonv3: So a thing I'm a little worried about--and I hadn't thought about this earlier--is that basically this could be used to just spam any user if they find out this endpoint. Right now we're constraining who has access to the API (the mobile app, the website) so it's less of a problem, but it might be a good idea to at least check the e-mail beforehand? 

I think all that has to happen in this PR for this is changing the middleware to use `userAuthenticated` (or whatever the function call is) and prevent non-authenticated users to use it. 

Then, in the mobile app, if someone requests purchase information, we can use the `verifyEmail` functionality to check the person owns the e-mail address? 

Is that overkill?

rann143: Would it be necessary to verify someone's email if they're already logged in? We could use that person's email attached to their profile, which I assume is already verified. This way we also wouldn't need to open another form on the app for them to fill out (the 'enter email' prompt).

Correct me if I'm wrong about the verifyEmail flow: They'd enter their email, receive a code, then input that code on the app to verify? If that's the case then it might be a bit overkill, especially if they already have a verified email connected to their account.

And in the case where someone is not logged in we could compel them to login/sign-up before getting the purchase info. The con to there is that it extends the purchase-cycle.

emails/album-purchase-link/html.pug
rann143: Implemented: [89a28ee](https://github.com/funmusicplace/mirlo/pull/1415/commits/89a28ee5ef5a88199d908d7febe48bd04e06a13e)

src/routers/v1/trackGroups/{id}/emailPurchaseLink.ts
simonv3: yep, I think if we require sign in beforehand we don't need to do the email verification step. It's probably easier that way. But I also feel like this might lend itself to a slow account creation step. 

"oh you wan to purchase, enter your email" -> creates an account. They verify the account. Then later we get them to set a password, but we log them in after they've verified the e-mail.

A problem for later though. I think right now the easiest is to just switch this endpoint to require authentication.

rann143: Sounds good.

Yeah for now I'm thinking if we're detecting whether the person's logged in or not - if yes, display the 'email info' button. Otherwise, prompt them to login/sign up to get purchase info. This way we avoid semi-account creation for the time being.

simonv3: Cool, can you use the `userAuthenticated` middleware in that case here rather than `userLoggedInWithoutRedirect`

rann143: Yup! Forgot to push that lol.

simonv3 (approved):

PR #1422, opened 2025-08-24
Fix Email Locals

rann143:
sendMail previously didn't have 'await' in front of it.

### DIFF
--- a/src/routers/v1/trackGroups/{id}/emailPurchaseLink.ts
+++ b/src/routers/v1/trackGroups/{id}/emailPurchaseLink.ts
@@ -36,8 +36,6 @@ export default function () {
           },
           locals: {
             trackGroup,
-            email,
-            host: process.env.API_DOMAIN,
             client: process.env.REACT_APP_CLIENT_DOMAIN,
           },
         },
--- a/src/routers/v1/trackGroups/{id}/emailPurchaseLink.ts
+++ b/src/routers/v1/trackGroups/{id}/emailPurchaseLink.ts
@@ -28,7 +28,7 @@ export default function () {
         res.status(404);
         return next();
       }
-      sendMail({
+      await sendMail({
         data: {
           template: "album-purchase-link",
           message: {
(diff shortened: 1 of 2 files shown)


src/routers/v1/trackGroups/{id}/emailPurchaseLink.ts
```diff
@@ -28,7 +28,7 @@ export default function () {
         res.status(404);
         return next();
       }
-      sendMail({
+      await sendMail({
```
simonv3: Was this causing issues? If not, do we want sendMail completing to block the return of this endpoint (ie is it fine if it's asynchronous?)

rann143: Yes your right, we wouldn't want it blocking the return. 

I'm having issues with the endpoint returning a 200 success but no email being sent. I'll take another look at the original addition that got merged.

simonv3: Is this happening locally? Or with the production API? I could take a look at sendgrid to see if an e-mail has gone missing? Or check the logs?

rann143: It was happening with the production api. Yes that would be super helpful if you see anything

simonv3: Oh, it turns out this was crashing the server? That might be why you weren't getting e-mails haha. I've got a fix incoming. [d36f598](https://github.com/funmusicplace/mirlo/pull/1428/commits/d36f5982566ff9dcf31aeea71622b0b2f06f43c3)

rann143: Ah true hah. Thanks! I also made another update in this PR on top of your change. Needed to update the other local variable in the subject.pug

I also removed unused locals & removed using await from this PR

PR #1503, opened 2025-09-22
assign processed title

### LINKED ISSUE #1213: Uploaded track titles (file name) don't recognize accents
Image
In this case the accent was on the file name and ended up not being recognized. If accents are not allowed, then replacing them with the unaccented name should be better.

MaciAC:
fixes #1213 

title was properly checked but not assigned


simonv3 (approved):
good catch

PR #1512, opened 2025-09-27
modify text editor toolbar behaviour

MaciAC:
set sticky behaviour with height taken from EditPostHeader component


simonv3 (approved):
looks great!

PR #1513, opened 2025-09-27
fix jump when insert url

### LINKED ISSUE #1491: For long blog posts, if you add a link, the page jumps up, but the input field for the URL stays put
The page shouldn't jump up when trying to enter a URL in long blog posts

MaciAC:
fixes #1491


simonv3 (approved):
The quality of life improvement that this is can not be overstated. Truly thank you.

PR #1520, opened 2025-10-01
Localize payment slider modal content

### LINKED ISSUE #1475: Most text in "How much goes to Mirlo?" window isn't translatble
Image
I'm talking about the little window that pops up when you lcick on the question mark when choosing well, how much goes to Mirlo with that release.

neuralsorcerer:
Changes:
- Localized the payment slider modal content by switching the title, explanatory copy, tier table, and finance link to translated strings rendered with `<Trans>`, enabling full internationalization of the help popover.
- Added English translation entries for the modal text, calculator paragraph, tier breakdown data, and finance link so the new strings can be translated via Transifex.

Why?
- Closes #1475


client/src/translation/en.json
```diff
@@ -660,7 +660,31 @@
     "fullCost": "Full cost",
     "platformCut": "Platform cut: {{percent}}%",
     "redistribution": "Redistribution",
-    "howMuchModalText": "Mirlo thrives on artist contributions—whether through monthly support or a slice of album sales. But we get it: every cent counts, and only you know what you can afford. That’s why we let you decide how much of your sales goes to Mirlo. Your art, your call."
+    "howMuchModalText": "Mirlo thrives on artist contributions—whether through monthly support or a slice of album sales. But we get it: every cent counts, and only you know what you can afford. That's why we let you decide how much of your sales goes to Mirlo. Your art, your call.",
+    "howMuchModalCalculator": "If you're having trouble deciding on a rate, you can use <Link>this sliding scale calculator</Link> for individuals provided by the Aorta Co-operative.",
+    "howMuchModalTableRows": [
```
simonv3: Ah, unfortunately I don't think Transifex supports nested things like this. We've had issues with it in the past.

simonv3:
This is great! But I'm not sure this will work with how transifex handles our translations. I don't think it handles arrays / lists, and iirc it struggles with more nested content (something we tried in the past)

client/src/translation/en.json
neuralsorcerer: ah, you're correct. updated. wrapped inside `useMemo`

simonv3 (approved):

neuralsorcerer:
> This is great! But I'm not sure this will work with how transifex handles our translations. I don't think it handles arrays / lists, and iirc it struggles with more nested content (something we tried in the past)

Replaced the modal’s linked paragraphs with flat translation segments so the help content no longer depends on nested <Trans> markup.

PR #1522, opened 2025-10-02
Localize SmallTileDetails labels

neuralsorcerer:
Changes:
- Localized the SmallTileDetails subtitle and footer using `Trans`, so translators can control the full “from/by” phrases shown with release metadata.
- Added English translation templates for the SmallTileDetails subtitle and footer strings to support localization work.

### DIFF
--- a/client/src/components/common/SmallTileDetails.tsx
+++ b/client/src/components/common/SmallTileDetails.tsx
@@ -2,6 +2,7 @@ import { css } from "@emotion/css";
 import { bp } from "../../constants";
 import React from "react";
 import styled from "@emotion/styled";
+import { Trans } from "react-i18next";
 
 const WidgetSection = styled.div`
   overflow: hidden;
@@ -28,6 +29,10 @@ export const SmallTileDetails: React.FC<{
   footer,
   textColor = "var(--mi-normal-foreground-color)",
 }) => {
+  const SubtitleComponent = () => <span>{subtitle}</span>;
+
+  const FooterComponent = () => <>{footer}</>;
+
   return (
     <>
       <div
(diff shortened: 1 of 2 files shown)


client/src/components/common/SmallTileDetails.tsx
```diff
@@ -28,6 +29,10 @@ export const SmallTileDetails: React.FC<{
   footer,
   textColor = "var(--mi-normal-foreground-color)",
 }) => {
+  const SubtitleComponent = () => <span>{subtitle}</span>;
```
simonv3, line 32: I'd prefer these in the view itself, I think it's a little confusing when we do that kind of stuff outside of the view.

simonv3: Oops accidsentally deleted a comment here. 

I think it's easier to track if we put this in the Trans component itself. Is there a reason for making it a function like this?

client/src/components/common/SmallTileDetails.tsx
```diff
@@ -83,8 +88,14 @@ export const SmallTileDetails: React.FC<{
               }
             `}
           >
-            <span>from </span>
-            <span>{subtitle}</span>
+            <Trans
```
simonv3, line 87: Oh, this definitely saves some code with not having to import the hook. Nice.

client/src/components/common/SmallTileDetails.tsx
neuralsorcerer: yup, you're correct. not a particular reason, didn't think about much while creating those helper comps, lol.

simonv3 (approved):

PR #1523, opened 2025-10-02
Limit loading state to active promo code button

### LINKED ISSUE #1480: "Download all" button wrongly animates when download only one code batch
Image

On the promo code page, when you click on 1 to download only a single batch of codes, the animation for the Download All button triggers, but it shouldn't. Instead, the button that was clicked on itself should probably animate.

neuralsorcerer:
Changes:
- Tracked the active promo-code download target (all codes vs. a specific group) and reset it after each request so only the relevant control shows its loading spinner.
- Passed a per-row `isDownloading` flag into `AlbumCodesRow` and bound the CSV button’s `isLoading` prop to that flag, ensuring only the clicked row animates during its download.

Why?
- Closes #1480


simonv3 (approved):
lgtm

PR #1524, opened 2025-10-02
Invert album availability toggle and copy

### LINKED ISSUE #1216: Reverse description of `manageAlbum.isGettable`
Following #1206 i think `manageAlbum.isGettable` should be changed to streamline the process of uploading albums.
Instead of "Is it possible for a user to buy or download this album?" - a question that's ticked as correct by default - this be unticked by default and read "Prevent users from downloading and buying this album". We should aim to have most default settings as unticked by default, so that the person uploading an album feels like they're choosing something rather than fighting against a default which chose things in advance for them.
The follwoing keys `manageAlbum.isGettableTooltip` and ` manageAlbum.isGettableInfo` should then be modified as a result.

neuralsorcerer:
Changes:
- Inverted the manage-album checkbox default to represent “prevent purchases” in the form while continuing to show pricing controls only when the album remains available.
- Extended `SavingInput` with an optional value transformer so checkbox values can be flipped before persisting to the API.
- Rewrote the manage-album tooltip and helper text to stress preventing downloads and purchases.

Why?
- Closes #1216


client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/SavingInput.tsx
```diff
@@ -70,6 +72,10 @@ const SavingInput: React.FC<{
         value = value ? value * 100 : undefined;
       }
 
+      if (valueTransform) {
```
simonv3, line 75: I like this

PR #1529, opened 2025-10-02
Add newsletter signup form and API

### LINKED ISSUE #1167: Add newsletter sign-up to the main page
This means adding a form on the main page that will insert create a blank user and add them to the "instanceArtist follower list.

neuralsorcerer:
Changes:
- Added a styled newsletter sign-up section to the homepage, wired it into the layout, and provided the necessary translations for the new copy and form labels.
- Implemented a `/v1/newsletter` endpoint that validates email input, creates or updates a listener account, and follows the instance artist so newsletter subscribers are tracked, and registered the route with the API router list.
- Added API coverage to confirm the newsletter workflow creates new listeners or updates existing ones while subscribing them to the instance artist.

Why?
- Closes #1167

image

### DIFF
--- a/src/routers/v1/artists/{id}/follow.ts
+++ b/src/routers/v1/artists/{id}/follow.ts
@@ -1,6 +1,5 @@
 import { User } from "@mirlo/prisma/client";
 import { NextFunction, Request, Response } from "express";
-import fetch from "node-fetch";
 
 import { userLoggedInWithoutRedirect } from "../../../../auth/passport";
 import prisma from "@mirlo/prisma";
@@ -12,43 +11,9 @@ import {
 } from "../../../../utils/artist";
 import { AppError } from "../../../../utils/error";
 import { getSiteSettings } from "../../../../utils/settings";
+import { checkCloudFlareTurnstile } from "../../../../utils/cloudflare";
 
 const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
-const { CLOUDFLARE_TURNSTILE_API_SECRET } = process.env;
-
-async function checkCloudFlare(token: unknown, ip: string | undefined) {
-  if (!CLOUDFLARE_TURNSTILE_API_SECRET) {
-    return;
-  }
-
-  if (!token || typeof token !== "string") {
-    throw new AppError({
-      httpCode: 400,
-      description: "Spam protection challenge is required",
-    });
-  }
-
-  const url = "https://challenges.cloudflare.com/turnstile/v0/siteverify";
-  const result = await fetch(url, {
-    body: JSON.stringify({
-      secret: CLOUDFLARE_TURNSTILE_API_SECRET,
-      response: token,
-      remoteip: ip,
-    }),
-    method: "POST",
-    headers: {
-      "Content-Type": "application/json",
-    },
-  });
-
-  const outcome = await result.json();
-  if (!outcome.success) {
-    throw new AppError({
-      httpCode: 400,
-      description: "Spam protection failed",
-    });
-  }
-}
 
 type Params = {
   id: string;
@@ -87,7 +52,11 @@ export default function () {
       if (!isLoggedIn) {
         const connectingIP =
           req.get("cf-connecting-ip") ?? req.ip ?? req.socket.remoteAddress;
-        await checkCloudFlare(cfTurnstile, connectingIP ?? undefined);
+        await checkCloudFlareTurnstile({
+          token: cfTurnstile,
+          ip: connectingIP ?? undefined,
+          skipIfNoSecret: true,
+        });
       }
 
       const artist = await prisma.artist.findFirst({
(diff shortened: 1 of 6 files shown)


src/routers/v1/newsletter.ts
```diff
@@ -0,0 +1,125 @@
+import { NextFunction, Request, Response } from "express";
```
simonv3: This is really great!

I'm wondering--to prevent spam abuse--if we need to verify these e-mails if the user hasn't been verified before (we can use the [code sending functionality](https://github.com/funmusicplace/mirlo/blob/main/client/src/components/common/EmailVerification.tsx#L14) we have), and / or, if we want to use the cloudflare functionality we have for non-logged in users. You can see it [on the flag endpoint](https://github.com/funmusicplace/mirlo/blob/main/src/routers/v1/flag.ts#L8) and [on the front-end](https://github.com/funmusicplace/mirlo/blob/main/client/src/components/TrackGroup/FlagContent.tsx#L40).

What do you think?

simonv3 (changes requested):
Wondering about your thoughts on preventing spam sign ups.

src/routers/v1/newsletter.ts
neuralsorcerer: Thank you for pointing that out. I completely forgot to account for spam abuse while implementing this.

Here are some recent updates:
- Updated the homepage newsletter form to require a verified email address, hook into Cloudflare Turnstile, and surface new helper copy guiding the verification step.
- Extended the reusable email verification component so callers can control its email field and have it reset state when the address changes, keeping the newsletter input and verification flow in sync.
- Hardened the newsletter API by validating Turnstile tokens when configured, insisting on pre-verified users, and clearing email confirmation tokens when codes are confirmed so subscribers are marked verified.

neuralsorcerer: Thank you for pointing that out. I completely forgot to account for spam abuse while implementing this.

Implemented verification.

src/routers/v1/newsletter.ts
```diff
@@ -0,0 +1,214 @@
...
+    });
+  }
+}
+
+export default function () {
+  const operations = {
+    POST: [POST],
+  };
+
+  async function POST(req: Request, res: Response, next: NextFunction) {
```
simonv3: Sorry this comes so late in the review, at first I didn't think too much about it, but now I'm wondering--any reason you don't think this should just be rolled in to the `artist/:id/follow` endpoint?

simonv3: You could get the artistId from /settings/instanceArtist, and then this would also round out some of the work for making `/artist/:id/follow` usable when not logged in.

neuralsorcerer: Updated.

src/routers/v1/artists/{id}/follow.ts
```diff
@@ -9,6 +10,45 @@ import {
   createSubscriptionConfirmation,
   subscribeUserToArtist,
 } from "../../../../utils/artist";
+import { AppError } from "../../../../utils/error";
+import { getSiteSettings } from "../../../../utils/settings";
+
+const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+const { CLOUDFLARE_TURNSTILE_API_SECRET } = process.env;
+
+async function checkCloudFlare(token: unknown, ip: string | undefined) {
```
simonv3: Could this re-use the function already defined in `v1/flags`?

src/routers/v1/artists/{id}/follow.ts
```diff
@@ -36,26 +100,100 @@ export default function () {
...
-      } else {
-        const user = await prisma.user.findFirst({
-          where: {
-            id: loggedInuser.id,
-          },
+      }
+
+      const settings = await getSiteSettings();
+      const instanceArtistId = settings.settings?.instanceArtistId;
+      const requiresVerifiedEmail =
```
simonv3: I might be missing something here, but why don't we want to require a verified e-mail for an other artist?

simonv3 (commented):
Looking good! Just two comments, curious to your thoughts. 

Again, massive thanks.

src/routers/v1/artists/{id}/follow.ts
neuralsorcerer: Centralized the Cloudflare Turnstile verification into `checkCloudFlareTurnstile` so multiple endpoints share the same validation and error handling.

src/routers/v1/artists/{id}/follow.ts
neuralsorcerer: Updated by requiring a verified email for every artist follow while only auto-opting instance-artist followers into the mailing list, preventing unverified listeners from subscribing.

neuralsorcerer:
> Looking good! Just two comments, curious to your thoughts.
> 
> Again, massive thanks.

Updated with changes and thank you so much for your reviews.

PR #1533, opened 2025-10-02
detailed description

Cwpute:


Cwpute:
so uh, this should have been on Cwpute-branch-1 but that branch did not seem up to date somehow ? i don't know what to do @simonv3 😬

PR #1539, opened 2025-10-04
Handle pasted images by uploading to post storage

### LINKED ISSUE #1489: When someone copy pastes text with images included, the images don't get parsed as images but instead become data/xml
This causes our memory usage to be way too big when fetching 1000+ of them from the database. 

When a user copy pastes text into the post portal, we need to parse images and upload them.

neuralsorcerer:
Changes:
- Added client-side handling for pasted data URI images in the post text editor, converting them into uploaded files, replacing inline sources with hosted URLs, and refreshing the post image list after successful uploads.
- Also added tracking for in-progress pasted image uploads and reuse of cached results so duplicate data URI uploads are skipped while conversions complete.

Why?
- Fixes #1489

### DIFF
--- a/client/src/components/common/TextEditor/index.tsx
+++ b/client/src/components/common/TextEditor/index.tsx
@@ -63,6 +63,128 @@ const TextEditor: React.FC<{
   });
 
   const [headerHeight, setHeaderHeight] = useState(0);
+  const processedPastedImagesRef = useRef<Map<string, string>>(new Map());
+  const uploadingPastedImagesRef = useRef<Set<string>>(new Set());
+
+  const dataUriToFile = useCallback((dataUri: string, index: number) => {
+    const [meta, base64Data] = dataUri.split(",");
+    const mimeMatch = meta?.match(/data:(.*?);/);
+    const mimeType = mimeMatch?.[1] ?? "image/png";
+
+    const extensionMap: Record<string, string> = {
+      "image/jpeg": "jpg",
+      "image/jpg": "jpg",
+      "image/png": "png",
+      "image/gif": "gif",
+      "image/webp": "webp",
+      "image/svg+xml": "svg",
+    };
+
+    const extension = extensionMap[mimeType] ?? "png";
+
+    const byteString = atob(base64Data ?? "");
+    const arrayBuffer = new Uint8Array(byteString.length);
+    for (let i = 0; i < byteString.length; i++) {
+      arrayBuffer[i] = byteString.charCodeAt(i);
+    }
+
+    return new File([arrayBuffer], `pasted-image-${Date.now()}-${index}.${extension}`, {
+      type: mimeType,
+    });
+  }, []);
+
+  const uploadPastedImages = useCallback(
+    async (html: string) => {
+      if (!postId) {
+        return;
+      }
+
+      if (!html.includes("data:image")) {
+        return;
+      }
+
+      const parser = document.createElement("div");
+      parser.innerHTML = html;
+
+      const images = Array.from(parser.querySelectorAll<HTMLImageElement>("img"));
+      const replacements: { original: string; replacement: string }[] = [];
+      let shouldReloadImages = false;
+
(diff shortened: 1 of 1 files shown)


client/src/components/common/TextEditor/index.tsx
```diff
@@ -173,8 +295,10 @@ const TextEditor: React.FC<{
         manager={manager}
         state={state}
         onChange={(parameter) => {
-          onChange(prosemirrorNodeToHtml(parameter.state.doc));
+          const html = prosemirrorNodeToHtml(parameter.state.doc);
+          onChange(html);
           setState(parameter.state);
+          void uploadPastedImages(html);
```
simonv3: I'm a little worried that as it is, this might overwrite a user's state after upload is complete if they moved on to other things? 

Is there a way to intercept an "on paste" event?

client/src/components/common/TextEditor/index.tsx
```diff
@@ -63,6 +63,128 @@ const TextEditor: React.FC<{
...
+    for (let i = 0; i < byteString.length; i++) {
+      arrayBuffer[i] = byteString.charCodeAt(i);
+    }
+
+    return new File([arrayBuffer], `pasted-image-${Date.now()}-${index}.${extension}`, {
+      type: mimeType,
+    });
+  }, []);
+
+  const uploadPastedImages = useCallback(
```
simonv3: This is great! Wondering if the majority (just the parsing bit) of it could be moved to an external function? Or as a hook?

client/src/components/common/TextEditor/index.tsx
neuralsorcerer: Great catch! Thanks

client/src/components/common/TextEditor/index.tsx
neuralsorcerer: Updated

client/src/components/common/TextEditor/index.tsx
```diff
@@ -22,6 +22,235 @@ import TopToolbar from "./TopToolbar";
...
+}: UsePastedImageUploadArgs) => {
+  const processedPastedImagesRef = useRef<Map<string, string>>(new Map());
+  const uploadingPastedImagesRef = useRef<Set<string>>(new Set());
+
+  useEffect(() => {
+    processedPastedImagesRef.current.clear();
+    uploadingPastedImagesRef.current.clear();
+  }, [postId]);
+
+  const dataUriToFile = useCallback(async (dataUri: string, index: number) => {
```
simonv3: If this doesn't rely on data outside of the function itself, could we define it in the parent scope?

client/src/components/common/TextEditor/index.tsx
```diff
@@ -22,6 +22,235 @@ import TopToolbar from "./TopToolbar";
...
 
+type UsePastedImageUploadArgs = {
+  manager: { view: any };
+  postId?: number;
+  reloadImages?: () => void;
+  setState: ReturnType<typeof useRemirror>["setState"];
+  onChange: (html: string) => void;
+};
+
+export const usePastedImageUpload = ({
```
simonv3: Could this be put in its own file?

neuralsorcerer: Updated.

client/src/components/common/TextEditor/usePastedImageUpload.test.tsx
```diff
@@ -0,0 +1,334 @@
...
+    expect(api.uploadFile).toHaveBeenCalledWith("manage/posts/123/images", expect.any(Array));
+    expect(onChange).not.toHaveBeenCalled();
+    expect(setState).not.toHaveBeenCalled();
+    expect(reloadImages).not.toHaveBeenCalled();
+
+    unmount();
+    destroy();
+  });
+
+  test("retains later edits while replacing image sources", async () => {
```
simonv3, line 284: 👍🏻

PR #1540, opened 2025-10-04
Handle cancelled Stripe checkout redirects

### LINKED ISSUE #907: Handle cancelled purchase response from Stripe
Right now we just display json on the API, but we should redirect to the front-end and describe what went wrong

neuralsorcerer:
Changes:
- Redirected cancelled Stripe checkout requests to a frontend error page, using client and artist lookups to choose the appropriate destination and providing a reason code for display.
- Added client and artist identifiers to Stripe cancel URLs for catalogue purchases, tips, and subscriptions so the API can generate accurate redirects after a cancellation.
- Created a checkout error view with routing and translations to explain cancelled purchases to users on both artist-specific and global pages.

Why?
- Fixes #907

### DIFF
--- a/client/src/components/Checkout/CheckoutError.tsx
+++ b/client/src/components/Checkout/CheckoutError.tsx
@@ -1,0 +1,77 @@
+import { css } from "@emotion/css";
+import { useMemo } from "react";
+import { useTranslation } from "react-i18next";
+import { Link, useParams, useSearchParams } from "react-router-dom";
+
+import { WidthWrapper } from "components/common/WidthContainer";
+
+type MessageKey =
+  | "generic"
+  | "userCanceled"
+  | "abandoned"
+  | "paymentFailed"
+  | "expired";
+
+const REASON_TRANSLATION_MAP: Record<string, MessageKey> = {
+  user_canceled: "userCanceled",
+  abandoned: "abandoned",
+  expired: "expired",
+  failed: "paymentFailed",
+  failed_invoice: "paymentFailed",
+  void_invoice: "paymentFailed",
+  declined: "paymentFailed",
+  duplicate: "paymentFailed",
+  fraudulent: "paymentFailed",
+  expired_card: "paymentFailed",
+  lost_card: "paymentFailed",
+  stolen_card: "paymentFailed",
+  testmode_payment_canceled: "paymentFailed",
+};
+
+const DEFAULT_MESSAGE_KEY: MessageKey = "generic";
+
+type RouteParams = {
+  artistId?: string;
+};
+
+function CheckoutError() {
+  const { t } = useTranslation("translation", { keyPrefix: "checkoutError" });
+  const [searchParams] = useSearchParams();
+  const { artistId } = useParams<RouteParams>();
+
+  const reason = (searchParams.get("reason") ?? "").toLowerCase();
+  const messageKey = REASON_TRANSLATION_MAP[reason] ?? DEFAULT_MESSAGE_KEY;
+  const homePath = useMemo(
+    () => (artistId ? `/${encodeURIComponent(artistId)}` : "/"),
+    [artistId]
+  );
+
+  return (
+    <WidthWrapper
+      variant="small"
+      className={css`
+        margin-top: 4rem !important;
+        text-align: center;
+
+        h1 {
+          margin-bottom: 1.5rem;
+        }
+
+        p {
+          margin-bottom: 2rem;
+        }
+
+        a {
+          color: var(--color-primary);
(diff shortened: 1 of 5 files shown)


neuralsorcerer (commented):
In monthly income report job `src/jobs/send-out-monthy-income-report.ts`

- Aggregated the monthly income report by each artist owner, caching user records, skipping owners without deliverable emails or zero revenue, and normalizing sale amounts before dispatching the mail job so multi-artist owners receive a single consolidated email.

client/src/components/Checkout/CheckoutError.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+        a {
+          color: var(--color-primary);
+          font-weight: 600;
+        }
+      `}
+    >
+      <h1>{t("title")}</h1>
+      <p>{t(messageKey)}</p>
+      <Link to="/">{t("returnHome")}</Link>
```
simonv3: Do we need this on an artist page?

src/routers/v1/checkout/index.ts
```diff
@@ -3,14 +3,102 @@ import prisma from "@mirlo/prisma";
...
 
   async function GET(req: Request, res: Response, next: NextFunction) {
-    const { success, session_id, stripeAccountId } = req.query;
+    const { success, canceled, session_id, stripeAccountId } = req.query;
     try {
+      if (canceled) {
+        const [clientId, artistId] = [
+          parseNumericQueryParam(req.query.clientId),
+          parseNumericQueryParam(req.query.artistId),
+        ];
```
simonv3: This feels like it would be much easier to read like this no?
```suggestion
        const clientId = parseNumericQueryParam(req.query.clientId);
        const artistId = parseNumericQueryParam(req.query.artistId);
```

src/routers/v1/checkout/index.ts
```diff
@@ -3,14 +3,102 @@ import prisma from "@mirlo/prisma";
...
+
+        const [client, artist] = await Promise.all([
```
simonv3: Ditto

src/jobs/send-out-monthy-income-report.ts
```diff
@@ -29,27 +29,92 @@ const sendOutMonthlyIncomeReport = async () => {
...
+      const user = usersById.get(userId);
+
+      if (!user?.email) {
+        logger.warn(
+          `Skipping monthly income report for artist owner ${userId} because no email address was found.`
+        );
+        continue;
+      }
+
+      const normalizedSales = userSales.map((sale) => ({
```
simonv3: I don't think amount can be anything other than a number because it comes from the database. So this feels unnecessary to me.

src/jobs/send-out-monthy-income-report.ts
```diff
@@ -29,27 +29,92 @@ const sendOutMonthlyIncomeReport = async () => {
       orderBy: { datePurchased: "asc" },
     });
 
-    const mappedArtists = keyBy(allArtists, "id");
```
simonv3, line 32: It's not clear to me what this does that wasn't already happening? It was already grouping all sales for a single user and it feels much more verbose and harder to track.

src/routers/v1/checkout/index.ts
```diff
@@ -38,15 +124,14 @@ export default function () {
           tipId: string | null;
         };
         if (clientId && artistId) {
-          const client = await prisma.client.findFirst({
-            where: {
-              id: +clientId,
-            },
+          const clientPromise = prisma.client.findUnique({
```
simonv3: could we `await` these directly?

src/routers/v1/checkout/index.ts
```diff
@@ -3,14 +3,100 @@ import prisma from "@mirlo/prisma";
...
 
   async function GET(req: Request, res: Response, next: NextFunction) {
-    const { success, session_id, stripeAccountId } = req.query;
+    const { success, canceled, session_id, stripeAccountId } = req.query;
     try {
+      if (canceled) {
+        const clientId = parseNumericQueryParam(req.query.clientId);
+        const artistId = parseNumericQueryParam(req.query.artistId);
+
+        const clientPromise = clientId
```
simonv3: Could we await these directly?

src/routers/v1/checkout/index.ts
neuralsorcerer: Updated

src/routers/v1/checkout/index.ts
neuralsorcerer: Updated.

PR #1541, opened 2025-10-04
Require processed tracks before publishing

### LINKED ISSUE #981: Prevent users from publishing incomplete releases
> preventing people from pressing publish if they don't have successful tracks or a cover

neuralsorcerer:
Changes:
- Make sure the publish to require at least one track with a successfully uploaded audio file before allowing publication.
- Added appropiate tests and validations.

Why?
- Fixes #981


simonv3:
Agreed, it's very good to have tests!

I'd like to flesh out our E2E tests at some point too, but we haven't properly set those up yet. Not sure if that's [something you want to take a stab at](https://github.com/funmusicplace/mirlo/issues/1547)? Either way, really appreciate the changes you've been putting in.

PR #1543, opened 2025-10-05
Allow resending email verification

### LINKED ISSUE #917: Sign up flow: allow user to resend verification e-mail
> Trying to set up a Mirlo account. The registration email took super long to arrive. By the time it did the token had expired. Now I can't log-in nor can I reset my password. If I try to sign-up again it says there is already a user with that email (which is true...)

Basically proposing that when the user tries to sign up, and they get the "user already exists" email, to pop up an option that says "resend verification email" if the user email hasn't been verified (ie if there's still a token).

* [ ] UI logic needs to check for token before displaying the option
* [ ] Will need its own endpoint to initiate the verification flow, this will need to send an email to the user
* [ ] Can probably re-use the second part of the verification flow (the link they click from the email).

neuralsorcerer:
Changes:
- Updated the signup route to detect existing accounts awaiting verification and return structured guidance with expiration details.
- Added a `/auth/resend-verification-email` endpoint that regenerates confirmation tokens and reuses the existing email flow.
- Extracted a reusable verification-email helper that normalizes account types and centralizes the notification dispatch.

Why?
- Closes #917


simonv3 (commented):
This looks good, but I'm not sure it closes the ticket right--wouldn't a user need to request the resending on the front-end?

simonv3 (approved):

PR #1545, opened 2025-10-05
fix: export more messages for translation

nykula:
When testing my translation (apart from notes at #1530), I found quite a few strings that weren't made translatable or had missing or wrong keys. I fixed them on my test instance and would like to share the fixes. Here are the pages where I found the strings, and the strings themselves.

```
manage/artists/1/post/1/
+ Type something
+ featured image

test/posts/mij-dopis/
+ by Тест

manage/artists/1/release/2
+ Uploading {filename}

manage/artists/1/release/2
(List them in the track listing)
+ Remove this artist

manage/artists/1/release/2
(Add license dialog)
+ add

manage/artists/1/release/2
+ Show Raw ID3 Data

/hmm/release/mi-temp-slug-new-album-9aabb6ba-f584-4f4c-a7ba-b0b1eb3b6c51
(where add to collection, before album is published)
(also in private mode)
+ Loading...

/hmm/release/mi-temp-slug-new-album-9aabb6ba-f584-4f4c-a7ba-b0b1eb3b6c51
(entire album)
+ Download

/hmm/release/mi-temp-slug-new-album-9aabb6ba-f584-4f4c-a7ba-b0b1eb3b6c51
(flag content)
+ Copyright violation
+ Inappropriate content

manage/artists/1/releases/tools
(promocode)
+ Success

profile/label
+ linkToProfile

profile/notifications
+ bought your album
+ New release by Тест
+ Admin followed your artist: Тест

manage/artists/2/customize
(labels/collectives)
+ No
+ Yes

hmm/release/mi-temp-slug-new-album-9aabb6ba-f584-4f4c-a7ba-b0b1eb3b6c51/download?email=admin@admin.example
+ Download your new release
+ Download

manage/artists/2/releases
(Set entire catalogue price)
+ Enter price

manage/artists/2/releases
(Dates)
(New tour date)
+ artist.remove

search?search=
+ releases.tags
+ releases.artists
+ releases.labels

manage/artists/1/tiers/supporters
+ Upload subscribers
+ Two ways to add emails
+ Next
```

Issue #1530, opened 2025-10-02
Notes on translation process


nykula:
Hello! While translating Mirlo into Ukrainian, I wrote a few notes that I would like to share.

- [ ] In [client/README.md](https://github.com/funmusicplace/mirlo/blob/42fe33e14a0b66a3076b2fa16e10b16582695dc3/client/README.md?plain=1#L5), instead of `REACT_APP_TRANSIFEX_TOKEN` the variable should be `VITE_TRANSIFEX_TOKEN`.

- [ ] The token isn't technically a secret (makes its way into the public build artifacts) so it might make sense to add it to `.env.example` and remove the note from about it from `client/README.md` so that people testing their work-in-progress translations locally can pull their recently translated messages from the cloud. I'm not sure about this, though, because I haven't worked with Transifex tokens much.

- [ ] Since the open-source application isn't complete without translated messages, they should be periodically backed up to the main version control repo (even if for now only the cloud version is used), for example like this:
```sh
for locale in $(
  sed -En 's,.*short: "(\w+)".*,\1,p' client/src/i18n.ts |grep -v en)
do
  curl -L "https://cds.svc.transifex.net/content/$locale" \
      --oauth2-bearer "$VITE_TRANSIFEX_TOKEN" \
    |jq '.data 
      |to_entries 
      |map(.key / "." + [.value.string]) 
      |group_by(.[0]) 
      |map({ key: .[0][0], value: map({ key: .[1], value: .[2] }) |from_entries })
      |from_entries' \
    >"client/src/translation/$locale.json"
done
```

- [ ] It might make sense to update `finishedLanguages` periodically based on how much is complete at a given moment. I see Polish and Dutch above 75% in the Transifex UI but they aren't in the list now.

- [ ] The order of strings in Transifex is quite chaotic, I have to consult [client/src/translation/en.json](https://raw.githubusercontent.com/funmusicplace/mirlo/42fe33e14a0b66a3076b2fa16e10b16582695dc3/client/src/translation/en.json) in another tab and search for e.g. `key=releases.` to work with a specific website section. If there's a way to reorder cloud strings programmatically to match the order in `en.json`, it would make the process more straightforward.

[...]

PR #1546, opened 2025-10-05
feat: slugify with rules of user language

nykula:
I like that Mirlo creates URL slugs using transliteration, this is an improvement over Bandcamp which, from what I remember, made `/--123` URLs for Cyrillic titles and rejected my support requests. However, during local testing of translations (#1530, #1545) I notice that Mirlo transliterates Ukrainian titles using Russian rules. For example `и` should become `y` but instead becomes `i`, e.g. for a blog post called `Мій допис` I want `mii-dopys` but have `mij-dopis`.

I read the documentation of `slugify`, the library used by Mirlo for transliteration, and notice that it offers a `locale` configuration parameter that turns on different rule presets depending on the language. I see a `language` field in the `User` entity on the Mirlo back end, and a `language` field in the `i18n` field returned by `useTranslation` on the front end, which, as far as I can see, can be passed to the `locale` of `slugify` as is, at least while the locales are named like `uk` and not `uk_UA`.

So, I patched my local back end to use this and I now get the `miy-dopys` transliteration for my language. `miy` still doesn't match the modern Ukrainian rules but a ruleset from 2000s is better than a ruleset of another language. I'm sharing this patch as a suggestion, but I have to note it was tested manually, I haven't figured out how to write a unit test for this project to guard against regressions.

### DIFF
--- a/client/src/components/common/SlugInput.tsx
+++ b/client/src/components/common/SlugInput.tsx
@@ -17,7 +17,9 @@ const SlugInput: React.FC<{
   currentName?: string;
 }> = ({ currentArtistId, isDisabled, type, currentName }) => {
   const { colors } = useGetArtistColors();
-  const { t } = useTranslation("translation", { keyPrefix: "manageArtist" });
+  const { i18n, t } = useTranslation("translation", {
+    keyPrefix: "manageArtist",
+  });
   const {
     register,
     watch,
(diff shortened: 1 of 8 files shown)


simonv3 (approved):
This is really great. 

For what it's worth there's documentation for back-end tests in `test/README.md` 

It's a little less documented but there's _some_ client side tests, which run using vitest. They're usually living next to the component being tested itself.

PR #1548, opened 2025-10-07
Fix supporter ordering

neuralsorcerer:
Changes:
- Added a sale-type priority map and enhanced the supporter sort logic so track group purchases stay ahead of tips when timestamps collide, keeping the endpoint’s ordering deterministic.

Why?
- Timestamps colliding while getting multiple supporter sources at the same time.

Check [this job.](https://github.com/funmusicplace/mirlo/actions/runs/18315088394/job/52153075469) Silent bug 💀

### DIFF
--- a/src/routers/v1/artists/{id}/supporters.ts
+++ b/src/routers/v1/artists/{id}/supporters.ts
@@ -18,6 +18,14 @@ const constructDateFilter = (
   }
 };
 
+const SALE_TYPE_ORDER: Record<string, number> = {
+  trackGroup: 0,
+  track: 1,
+  merch: 2,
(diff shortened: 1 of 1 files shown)


simonv3:
Ah, I assumed this was because the purchase time being generated by the test didn't properly fall in the window for the sales, since the test was randomly generating things. 

I'm fairly sure that `datePurchased` is guaranteed by the records in the database (it's required and sets a default `now()`), so I don't know that we need to do all the "this exists" checks?

neuralsorcerer:
Shall we keep these "good to have" checks? What do you think?

PR #1551, opened 2025-10-09
Allow artists to customize thank-you messages

### LINKED ISSUE #944: Thank you for your support / purchase text in subscription or purchase emails
The artist should be able to customize a part of the text in the email that gets sent to the customer. 

Ideally this would be basic html (similar to what's in the blog posts).

neuralsorcerer:
Changes:
- Added subscription management editors so artists can craft and save HTML thank-you messages for supporters and purchasers, persisting through artist properties, translations, and types.
- Updated subscription, tip, download, and purchase email templates to render an artist's custom message when available, falling back to the default closing text otherwise.

Why?
- Closes #944

### DIFF
--- a/client/src/components/ManageArtist/ManageArtistSubscriptionTiers.tsx
+++ b/client/src/components/ManageArtist/ManageArtistSubscriptionTiers.tsx
@@ -2,16 +2,18 @@ import { css } from "@emotion/css";
 import React from "react";
 import ManageSubscriptionTierBox from "./ManageSubscriptionTierBox";
 import SubscriptionForm from "./SubscriptionForm";
-import { useOutletContext } from "react-router-dom";
+import { useParams } from "react-router-dom";
 import SpaceBetweenDiv from "components/common/SpaceBetweenDiv";
 import { ManageSectionWrapper } from "./ManageSectionWrapper";
 import Modal from "components/common/Modal";
 import { useTranslation } from "react-i18next";
 import Button, { ButtonLink } from "components/common/Button";
 import { FaPlus, FaWrench } from "react-icons/fa";
-import { queryManagedArtistSubscriptionTiers } from "queries";
+import {
+  queryManagedArtist,
+  queryManagedArtistSubscriptionTiers,
+} from "queries";
 import { useQuery } from "@tanstack/react-query";
-import { ManageArtistOutletContext } from "./ManageArtist";
 
 const ManageArtistSubscriptionTiers: React.FC<{}> = () => {
   const [addingNewTier, setAddingNewTier] = React.useState(false);
@@ -19,13 +21,17 @@ const ManageArtistSubscriptionTiers: React.FC<{}> = () => {
     keyPrefix: "subscriptionForm",
   });
 
-  const { artist } = useOutletContext<ManageArtistOutletContext>();
+  const { artistId } = useParams();
+
+  const { data: artist } = useQuery(
+    queryManagedArtist(Number(artistId))
+  );
 
   const { data: tiers, refetch: refetchTiers } = useQuery({
     ...queryManagedArtistSubscriptionTiers({
-      artistId: artist?.id,
+      artistId: Number(artistId),
     }),
-    enabled: Boolean(artist?.id),
+    enabled: Boolean(artistId),
   });
 
   if (!artist) {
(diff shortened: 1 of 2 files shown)


client/src/components/ManageArtist/ManageArtistSubscriptionTiers.tsx
```diff
@@ -22,75 +26,262 @@ const ManageArtistSubscriptionTiers: React.FC<{}> = () => {
...
+    return typeof emails.support === "string" ? emails.support : "";
+  }, [artist]);
+
+  const originalPurchaseMessage = React.useMemo(() => {
+    const emails = (
+      (artist?.properties as { emails?: { purchase?: string | null } })?.emails ??
+      {}
+    ) as { purchase?: string | null };
+    return typeof emails.purchase === "string" ? emails.purchase : "";
+  }, [artist]);
```
simonv3: Could these be extracted into a function for easier reading? `getEmailTextFromProperties('purchase')` or something like that?

client/src/components/ManageArtist/ManageArtistSubscriptionTiers.tsx
```diff
@@ -22,75 +26,262 @@ const ManageArtistSubscriptionTiers: React.FC<{}> = () => {
...
+
+  const [supportMessage, setSupportMessage] = React.useState<string>("");
+  const [purchaseMessage, setPurchaseMessage] = React.useState<string>("");
+
+  React.useEffect(() => {
+    setSupportMessage(originalSupportMessage ?? "");
+    setPurchaseMessage(originalPurchaseMessage ?? "");
+  }, [originalSupportMessage, originalPurchaseMessage]);
+
+  const isEditorContentEmpty = React.useCallback((value: string) => {
```
simonv3: Could this be defined outside of the controller function since it's not dependent on anything in it?

client/src/components/ManageArtist/ManageArtistSubscriptionTiers.tsx
```diff
@@ -22,75 +26,262 @@ const ManageArtistSubscriptionTiers: React.FC<{}> = () => {
...
+    const updatedProperties: NonNullable<Artist["properties"]> =
+      artist.properties
+        ? { ...artist.properties }
+        : ({} as NonNullable<Artist["properties"]>);
+
+    if (Object.keys(updatedEmails).length > 0) {
+      updatedProperties.emails = updatedEmails;
+    } else {
+      delete updatedProperties.emails;
+    }
```
simonv3: I wonder if all of this could be turned into a slightly more generic function that could be used in other parts of the app to update artist (and user) properties? Also wonder if there's lodash functions that could help make this a little more readable?

client/src/components/ManageArtist/ManageArtistSubscriptionTiers.tsx
```diff
@@ -22,8 +22,10 @@ const ManageArtistSubscriptionTiers: React.FC<{}> = () => {
   });
 
   const { artistId } = useParams();
-  const { data: artist } = useQuery(queryManagedArtist(Number(artistId)));
-  const { data: tiers, refetch } = useQuery(
+  const { data: artist } = useQuery(
```
simonv3: Do we still need this in this case?

neuralsorcerer: Updated.

client/src/components/ManageArtist/ManageArtistSubscriptionTiers.tsx
```diff
@@ -2,95 +2,96 @@ import { css } from "@emotion/css";
...
   });
 
-  const { artistId } = useParams();
-  const { data: artist } = useQuery(queryManagedArtist(Number(artistId)));
-  const { data: tiers, refetch } = useQuery(
-    queryManagedArtistSubscriptionTiers({
-      artistId: Number(artistId),
-    })
-  );
+  const { artist } = useOutletContext<ManageArtistOutletContext>();
```
simonv3: Ah, what I guess I would have done is just reference the `artistId` from the `useParams()` rather than passing it from the context.

neuralsorcerer: Updated.

PR #1567, opened 2025-10-15
Setup proper e2e tests using cypress

neuralsorcerer:
Attempt to setup cypress for e2e tests.

Furthur updates shall be made in future.

### DIFF
--- a/client/cypress/e2e/home.cy.ts
+++ b/client/cypress/e2e/home.cy.ts
@@ -4,6 +4,17 @@ import {
   TRACK_GROUP_EXAMPLE,
 } from "../../test/mocks";
 
+type SnackbarRecord = {
+  msg: string;
+  type?: "success" | "warning";
+};
+
+declare global {
+  interface Window {
+    __mirloSnackbars?: SnackbarRecord[];
+  }
+}
+
 const newsletterEmail = "listener@example.com";
 const verificationCode = "123456";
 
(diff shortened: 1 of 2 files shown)


client/cypress/e2e/home.cy.ts
```diff
@@ -0,0 +1,208 @@
...
+    ]);
+  });
+
+  it("renders curated sections with links to releases, posts, and tags", () => {
+    cy.contains("Join the beta test for mirlo's upcoming iOS app!").should(
+      "be.visible"
+    );
+    cy.contains("Join Here")
+      .should("be.visible")
+      .and("have.attr", "href", "https://testflight.apple.com/join/SjkPHPqp");
```
simonv3: Don't need to test this as it'll change regularly (unless we can think of a way to make it not hard coded?)

client/test/vitest-setup.ts
```diff
@@ -1,9 +1,117 @@
-import { expect, afterEach } from "vitest";
+import { expect, afterEach, vi } from "vitest";
 import { cleanup } from "@testing-library/react";
 import * as matchers from "@testing-library/jest-dom/matchers";
 
 expect.extend(matchers);
 
+Object.assign(import.meta.env, { VITE_API_DOMAIN: "http://localhost:3000" });
```
simonv3, line 7: Let's keep this for now, but what do you think about running it against a full environment instead of mocking the API? Did you try it? I'm not sure how feasible it is.

client/cypress/e2e/home.cy.ts
neuralsorcerer: Was thinking about not making it hard-coded but as of now couldn't find any solution. So we'll leave that for future work.

client/test/vitest-setup.ts
neuralsorcerer: Added a note for this.

PR #1568, opened 2025-10-15
Fix subscription tier button text spacing issue

### LINKED ISSUE #1559: If already subscribed, button concatenates both sentences
Image

Pink-red button here has its two sentences kissing each other with no dots inbetween 👨‍❤️‍💋‍👨

Ani07-05:
- Add space between subscription status and price text to prevent concatenation
- Fix translation syntax error in youAreFollowingThisArtist key

Fixes #1559

### DIFF
--- a/client/src/components/common/SupportArtistTiersForm.tsx
+++ b/client/src/components/common/SupportArtistTiersForm.tsx
@@ -133,7 +133,6 @@ const SupportArtistTiersForm: React.FC<{
           (value?.isDefaultTier
             ? t("youAreFollowingThisArtist", { artistName: artist.name })
             : t("youAreAlreadySubscribed"))}
-        {isSubscribedToCurrentTier && " "}
         {!value && !isSubscribedToCurrentTier && t("chooseToContinue")}
         {value?.id !== currentTier?.id &&
           value &&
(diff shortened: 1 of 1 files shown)


client/src/components/common/SupportArtistTiersForm.tsx
```diff
@@ -131,8 +131,9 @@ const SupportArtistTiersForm: React.FC<{
       >
         {isSubscribedToCurrentTier &&
           (value?.isDefaultTier
-            ? t("youAreFollowingThisArtist, { artistName: artist.name }")
+            ? t("youAreFollowingThisArtist", { artistName: artist.name })
             : t("youAreAlreadySubscribed"))}
+        {isSubscribedToCurrentTier && " "}
```
simonv3: I think part of the issue is that it's displaying both options. So I don't think it should show the following options.

So it shouldn't say both "You are already following." and "continueWIthPricing" in the same place.

Ani07-05: makes sense , tbf I did think of it but  judging from the comment I thought it was a spacing/concatenation issue, will fix it and make another PR

PR #1569, opened 2025-10-16
fix: use shopping cart icon for buy button on small album tiles

### LINKED ISSUE #1566: Buy button text on small album tile in collection
Hovering over an album in the wishlist while in english is perfectly fine.
Image

When the language is set to something else, the Buy button can get out of hand.
Image

I think i remember already reporting this ? or something similar maybe.
Anyway, this could easily be fixed with a symbol for the buy button (like there is one for the "Add to collection" button and "Download" and "Play" ones.

We could use this shopping cart icon, which i'm not found of but is easily recognizable and understood : 

Image

Or we could re-use the tip icon that we have on blog posts pages: 

Image

GooglyBlox:
Resolves #1566
The buy button text was overflowing on small album tiles when the language was set to anything other than English. Replace the text with a shopping cart icon when the button is in collapsed state (on album tiles) to ensure it always fits properly regardless of language.

### DIFF
--- a/client/src/components/TrackGroup/PurchaseAlbumModal.tsx
+++ b/client/src/components/TrackGroup/PurchaseAlbumModal.tsx
@@ -9,7 +9,7 @@ import useArtistQuery from "utils/useArtistQuery";
 import { FixedButton } from "components/common/FixedButton";
 import { useQuery } from "@tanstack/react-query";
 import { queryArtist } from "queries";
-import { IoCart } from "react-icons/io5";
+import CurrencyCoinIcon from "components/common/CurrencyCoinIcon";
 
 const PurchaseAlbumModal: React.FC<{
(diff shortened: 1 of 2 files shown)


GooglyBlox:
Can do! Should I use the Donate icon referenced in https://github.com/funmusicplace/mirlo/issues/1566#issuecomment-3412375531 as a base?

GooglyBlox:
image
image

This is what I've gone with so far, would be happy to adjust.

GooglyBlox:
> I think it works! Is it a toggle we can turn on and off (dependent on the size we intend the album art to be?) (I think defaulting to full text will still be better)

Yup! It's a toggle that turns on/off based on the size/context (via the collapse prop), and it defaults to showing full text when not collapsed.

simonv3 (approved):

PR #1605, opened 2025-10-28
trackPlays endpoint: Remove Requirement to be Logged in

rann143:
It look's like `{ ip: req.ip }` from `...(user ? { userId: user.id } : { ip: req.ip }),` can't be reached since a user is required to be logged in?

Was trying to register trackPlays as a logged out user but kept getting an error that the api call was failing. Thinking this might've been the issue

### DIFF
--- a/src/routers/v1/tracks/{id}/trackPlay.ts
+++ b/src/routers/v1/tracks/{id}/trackPlay.ts
@@ -21,7 +21,7 @@ import { AppError } from "../../../../utils/error";
 
 export default function () {
   const operations = {
-    GET: [userLoggedInWithoutRedirect, GET],
+    GET,
   };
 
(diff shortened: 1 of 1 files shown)


simonv3:
Hmm, log in without redirect shouldn't fail, it should add the user object if they are logged in. And we do want to know that in the endpoint, so I think removing that middleware isn't right. 

Were you able to reproduce it locally? If so, did the API show any errors?

rann143:
I'm getting a 524 timeout error anytime I try to GET that particular endpoint whether logged in or not.

Looks like: ``await api.get(`/v1/tracks/${track.id}/trackPlay`, {});``

'api' is set up the same as from the fetchWrapper in mirlo. A difference is that I use `/v1/`.

PR #1610, opened 2025-10-30
add more metadata to track and trackGroup descriptions

sugzone:
Adding more metadata to track and trackGroup descriptions so that more info appears in `[og:description]` meta tag. This helps other projects such as [freq](https://freq.social/about) process relevant metadata when displaying albums from Mirlo. It also will help anyone who attempts to build a script to [seed MusicBrainz entires](https://musicbrainz.org/doc/Development/Seeding) from Mirlo URLs.

Other metadata might also be helpful here, but this is a barebones pull request intended to facilitate Mirlo embeds on Freq.

### DIFF
--- a/src/parseIndex.ts
+++ b/src/parseIndex.ts
@@ -300,6 +300,7 @@ const parseIndex = async (pathname: string) => {
         include: {
           artist: true,
           cover: true,
+          tracks: true,
         },
       });
       if (tg && route[4] === "tracks") {


src/parseIndex.ts
```diff
@@ -329,9 +329,16 @@ const parseIndex = async (pathname: string) => {
         }
       } else if (tg) {
         const coverString = tg.cover?.url.find((u) => u.includes("x600"));
+        //render a text tracklist to display in the description
+        const tracklist = ''
+        for ( const [ index, value ] of tg.tracks.entries() ) {
```
simonv3, line 335: On line 300, you'll need to `include` tracks for this to work.

sugzone: Ah, makes sense. I added that to the tg call. Let me know if that looks ok?

simonv3: Ah, yeah, so I missed the `value.artist` thing below. 

Tracks have `trackArtists` when there's more than one artist, but otherwise the artist is actually tied to the `trackGroup`. 

This is probably worth extracting into a util so that it's more re-usable:

https://github.com/funmusicplace/mirlo/blob/5ac533bb4e5789f711b305f96a3cc96f689eb977/src/utils/stripe.ts#L341

PR #1612, opened 2025-10-31
Merch and Track Sales Changes

simonv3:
Our e-mails to artists weren't including accurate transaction information. We now store those things so this fixes that. 

This also changes our sales transactions for our Merch & Track sales to display properly. 

Also skipped processing if a track doesn't exist.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -2,7 +2,7 @@
   "name": "@mirlo/client",
   "version": "0.1.0",
   "private": true,
-  "type": "module",
+  "type": "commonjs",
   "dependencies": {
     "@dnd-kit/core": "^6.3.1",
     "@dnd-kit/sortable": "^10.0.0",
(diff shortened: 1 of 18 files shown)


client/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@mirlo/client",
   "version": "0.1.0",
   "private": true,
-  "type": "module",
+  "type": "commonjs",
```
timglorioso, line 5: Vite doesn't seem happy about this. Curious why it was needed? I get "No built client" at localhost:3000 either way

simonv3: If you're accessing the front-end via port 3000 you need to build the client (the server serves the static files just like in production). So `yarn client:build` (IIRC). Is that what you did and it's still returning that message? 

This was necessary for Cypress to be able to import utility functions I think? It's been a minute at this point though.

PR #1631, opened 2025-11-17
three bugfixes for email verification

rubzo:
- One relates to making sure users in the admin interface have the correct confirmation status.
- The other reports a different message if a user tries to login with an unverified account, compared to telling them the username/password is incorrect.

### DIFF
--- a/client/src/components/Admin/AdminManageUser.tsx
+++ b/client/src/components/Admin/AdminManageUser.tsx
@@ -25,7 +25,7 @@ const AdminManageUser = () => {
   }, [id]);
 
   const onConfirmationEmailClick = React.useCallback(async () => {
-    await api.post(`manage/users/${id}/confirmEmail`, {});
+    await api.post(`users/${id}/confirmEmail`, {});
     callback();
   }, [callback, id]);
 


rubzo:
Will address the failing test!

src/routers/auth/login.ts
```diff
@@ -30,9 +30,18 @@ const login = async (req: Request, res: Response, next: NextFunction) => {
         });
       }
     } else {
+      let desc = "Incorrect username or password";
+      const potentiallyUnverifiedUser = await prisma.user.findFirst({
+        where: {
+          email: email.toLowerCase(),
+        },
+      });
+      if (potentiallyUnverifiedUser.emailConfirmationToken) {
```
simonv3: Is there a security downside to being less generic about this? Dunno if that outweighs the bad UX experience though.

rubzo: Yeah it was on my mind as well, but up to you really. I mean I thought normally the UX is either:

- unverified accounts are logged in, but redirected to some kind of holding pen where they are then invited to verify (with option to send another email)
- unverified accounts can still log in, but their actions are extremely limited (depends on how much this makes sense)

simonv3: I think we should eventually implement the second one for sure. We've talked about (and have the bones for IIRC) trust levels in the app, but nothing that moves users between them.  

Yeah, I think for UX's sake we can do this?

rubzo: Okay. But let me rewrite it so that you actually only get told you are unverified if you have provided the correct password, as that isn't currently how it works.

PR #1632, opened 2025-11-18
user: when deleting users, scramble email and remove PII

rubzo:
Also add an admin task to allow us to do this to users that have been deleted in the past.

### DIFF
--- a/src/utils/user.ts
+++ b/src/utils/user.ts
@@ -5,6 +5,24 @@ import logger from "../logger";
 import { deleteArtist, deleteStripeSubscriptions } from "./artist";
 import countries from "./country-codes-currencies";
 
+import { v4 as uuid } from "uuid";
+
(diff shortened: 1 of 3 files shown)


rubzo:
Just putting this up for reference/comment right now. Still missing some tests.

src/utils/user.ts
```diff
@@ -21,6 +39,10 @@ export const deleteUser = async (userId: number) => {
       userId: userId,
     },
   });
+
+  // The User object will be retained, so we scramble the email and remove other
+  // personal identifiers.
+  await anonymiseDeletedUser(userId);
   await prisma.user.delete({ where: { id: userId } });
```
timglorioso, line 46: Might anonymization come after deletion, on the off chance deletion fails?

timglorioso:
This looks good to me. What kinds of tests do we need?

Issue #1654, opened 2025-12-09
Change name of `Artist` table to `Profile` table


simonv3:
Why? 

Because Artists and Labels act the same, it makes sense for the profile page to more generically named. 

Change required:

* [ ] A migration that changes the name of the table
* [ ] Updating all the places where `prisma.artist` gets called and changing it to `prisma.profile`.

Potential thought:

This opens up the option for users to also use the `profile` to be the public display. I don't know if we want that honestly.

simonv3:
> Supporting Profile

I also think the overlap is a little blurry, because like, I can imagine a Big Fan actually regularly writing blog posts, in which case we're back at the "is it just a profile". And maybe purchases (and a listen history) are just a different view? What if you follow a friend who's a listener but not an artist (functionality you can do on Bandcamp).

timglorioso:
Yeah there's overlap for sure. I've thought about supporters blogging too. Maybe the Publishing/Supporting split/naming isn't quite right. I think the heart of it is in the relationships: artist on label roster, supporter not on label roster, things like that. It's subtle. I'll have to dig up some diagrams I was working on a few months ago.

LouisLouiseKay:
To me it's really down to : There's a main Profile template (which is more a backend/our way to say "these are all sort of the same code entity) and then there's Profile Types : Which is more a user facing way to tell users "create a new Listener profile / Artist profile / Label profile". Under the hood these are variants of the profile entity with various features enabled or disabled, and they allow us to build type specific features.

timglorioso:
Yeah but what I'm saying is I don't think a single Profile is best. I'll have to articulate this better.

LouisLouiseKay:
But it's not a single profile, it's a single "type of front facing architecture" which ultimately, it always kind of is : public facing entity x of type y wants to show to the world a list of things. That's why I think it's pretty good that we're thinking of decoupling *account* from profile, which is more "legal entity x owns access to *all these types of public facing profiles*".

And then for me the relationships you're talking about are relationships between different profile types.

simonv3:
(we can agree to rename the Artist table to Profile though?)

timglorioso:
Yeah it's a good improvement on what we have

timglorioso:
Just to make sure my WIP branch isn't overcomplicating things, a couple considerations:

1. change other Artist* tables? e.g. ArtistAvatar
2. change API endpoints? 
3. change frontend usage? (outside of API, e.g. queries)

[...]

PR #1685, opened 2026-01-05
fix page name on Announcement description box

Cwpute:
change "profile page" to "Artist page" because profile page would be understood as the the page where you set your properties, language etc... while Artist page is already used in public communication to talk about the public-facing Artist page on which you can see releases and merch etc...
I'd also like to make it so "markdown" gives a link to a "basic markdown rules"[ like this one](https://markdowncheatsheet.net/) page for users who don't know what Markdown is but i'm not sure how to do it.

### DIFF
--- a/client/src/translation/en.json
+++ b/client/src/translation/en.json
@@ -855,7 +855,7 @@
     "announcement": "Write an announcement:",
     "editAnnouncement": "Edit announcement",
     "close": "Close",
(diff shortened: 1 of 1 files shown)


simonv3:
@Cwpute what do you think about the discussion here? https://github.com/funmusicplace/mirlo/issues/1654

PR #1706, opened 2026-01-13
Mobile fixes tweaks

LouisLouiseKay:


client/src/components/ManageArtist/ManageArtistAlbums.tsx
```diff
@@ -58,6 +58,8 @@ const ManageArtistAlbums: React.FC<{}> = () => {
         <div
           className={css`
```
simonv3, line 59: For this kind of stuff could you use tailwind going forward?

PR #1741, opened 2026-01-31
Fix FAQ blocks and typos

NorbiPeti:
Fixed some FAQ blocks being nested into each other that don't seem to be closely related, also fixed some typos and a missing question.


simonv3:
Amazing thank you!!!

PR #1742, opened 2026-02-01
Vitepress translation setup

Cwpute:
This is the setup done i think, at least for french. We'll have to add directories for each language separately. I can do that later but i'd like to try how things work out with the french translation first.
As i understand Vitepress will "simply" generate a new page from what i added here.
I'm not sure Transifex will automagically detect the new files though ? Maybe something is needed so translations for the docs are made available there.

I'll leave it as a draft for later work and research. Hopefully.


docs/.vitepress/config.mts
```diff
@@ -67,4 +67,17 @@ export default defineConfig({
...
     ],
   },
+  locales: {
+    root: {
+      label: 'English',
+      lang: 'en'
+    },
+    fr: {
+      label: 'French',
+      title: 'test title'
```
simonv3:
```suggestion
      title: 'test title',
```

simonv3:
Amazing @Cwpute we can merge it (after the little missing comma) and see what happens. 

Unfortunately these translations won't live in Transifex, but will live directly in the codebase (you'll have to mimic the file structure) but in a way that makes sense to me. It'll give the most leeway for customizing for language -- eg where paragraphs end, etc. I thought the large text in transifex was a little tedious.

PR #1780, opened 2026-03-04
Improve accessibility of navigation menu

### LINKED ISSUE #1368: Screenreader and keyboard support for search and navigation (a.k.a. hamburger menu)
Probably could include a Skip to Content link in this effort too.

timglorioso:
Close #1368.

Converted to Tailwind (except `styled` components) and upgraded package to get `inset` classes used to position `<dialog>`.

- [x] Test with Orca screenreader
- [x] Test with VoiceOver screenreader
- [x] Test with keyboard
- [x] Test in Firefox 140, 146
- [x] Test in Chrome
- [x] Test in Safari
- [x] Resolve TS/React not recognizing valid `command` HTML attribute
- [x] Consider Menu button background or not
- [x] Fix Close Menu button icon size on viewports `md` and up
- [x] [Reset focus on navigate](https://gomakethings.com/shifting-focus-on-route-change-with-react-router/) (thanks Chris!)

image


timglorioso:
Not even React 19 supports [Invoker Commands API](https://developer.mozilla.org/en-US/docs/Web/API/Invoker_Commands_API). See https://github.com/facebook/react/issues/32478 and https://github.com/facebook/react/pull/32480. One more reason to dump React. Resolving with `@ts-ignore`.

client/src/components/Header/Header.tsx
```diff
@@ -195,73 +185,52 @@ const Header = () => {
...
-              `}
+              className="block me-[.75rem] no-underline text-center hover:underline! color-white! &_svg:fill-white! bg-black! max-md:text-(--mi-font-size-xsmall)!"
             >
-              <p>{t("donateNow")}</p>
+              <p>{t("donateNow", { keyPrefix: "kickstarter" })}</p>
             </ButtonLink>
           )}
+          {!isLoggedIn && <LogInPopup />}
+          {isLoggedIn && (
+            <Button
```
simonv3, line 209: I'm wondering if it makes sense to roll the <Button> <Menu> into one component that's re-usable, so that we don't have to refs etc every time we use them together (eg. every three dot menu in the app right?)

timglorioso: Yeah, whenever I get around to updating DropdownMenu I'll think about that. The Menu here is for Header only.

Issue #1690, opened 2026-01-08
Upgrade Prisma


simonv3:
We're running on Prisma 5. We should upgrade to Prisma 6.

PR #1815, opened 2026-03-23
chore: upgrade prisma to v6

simonv3:
This ended up taking much less time than it took up space in my head.

V7 next. 

https://github.com/funmusicplace/mirlo/issues/1690

### DIFF
--- a/prisma/prisma.ts
+++ b/prisma/prisma.ts
@@ -2,7 +2,8 @@ import { Prisma, PrismaClient } from "./__generated__";
 
 const ENABLE_LOGGING = false;
 
-const prisma = new PrismaClient({
+// Create base client before extending it
+const baseClient = new PrismaClient({
   // The following controls logging of the database
   ...(ENABLE_LOGGING
     ? {
@@ -21,7 +22,7 @@ const prisma = new PrismaClient({
  */
 
 // @ts-ignore
-prisma.$on("query", (e) => {
+baseClient.$on("query", (e) => {
   // @ts-ignore
   let queryString = e.query;
   // @ts-ignore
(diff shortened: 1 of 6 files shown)


prisma/migrations/migration_lock.toml
```diff
@@ -1,3 +1,3 @@
 # Please do not edit this file manually
-# It should be added in your version-control system (i.e. Git)
-provider = "postgresql"
\ No newline at end of file
+# It should be added in your version-control system (e.g., Git)
```
timglorioso, line 2: The pedant in me loves whoever did this

prisma/prisma.ts
```diff
@@ -36,63 +36,54 @@ prisma.$on("query", (e) => {
...
+ * - Intercepts find/findMany to filter out soft-deleted records
  *
- * It does NOT work for nested `where`, `include`, or `select` things
+ * It does NOT cascade delete and does NOT work for nested `where`, `include`, or `select` things
  */
-prisma.$use(async (params, next) => {
-  const hasDeletedAtField = Prisma.dmmf.datamodel.models
-    .find((m) => m.name === params.model)
-    ?.fields.find((field) => field.name === "deletedAt");
+export default prisma.$extends({
```
simonv3: RIP I spoke too soon. This was a pain in the butt. $use was deprecated so we had to switch to $extends, but this had the cascading effect of making type issues too complicated for the TS parser. 

@timglorioso do you have an opinion on the `deletedAt` flag? Is it something we should actually have? Or should we be explicit in all the queries where it's needed? (we already need to be explicit in nested queries). Does this obfuscate things a bit too much?

timglorioso: I think always soft deleting is generally good. So while it's now implicit, the DX is a little cleaner and it's safe by default.

PR #1816, opened 2026-03-24
Improve accessibility of release form

timglorioso:
See commits for a summary of what's improved. Other fixes and improvements resulting from testing are listed below.

- [x] Test with keyboard in Firefox on Linux
  - [x] Fix pricing mode stealing focus on page load and on min price change
  - [x] Add UploadFiles component to replace UploadWrapper and UploadField used in BulkTrackUpload and DownloadableContent so that we: support keyboard use, improve visual and screenreader experiences by keeping essential labeling, and improve the API and code quality of those UI components
- [x] Test with Orca in Firefox on Linux
  - [x] Fix label association for suggested price number input
  - [x] Improve use of `output` element in PaymentSlider by using `aria-atomic="true"` to announce both percent and tier on each change (previously was not announcing percent when tier changed, e.g. you would hear 14% then Redistribution then 16%)
  - [x] Add label to cover art delete button

Testing and fixes continued: https://github.com/funmusicplace/mirlo/pull/1945


client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PriceAndSuch.tsx
```diff
@@ -201,16 +219,13 @@ const PriceAndSuch: React.FC<{
...
         <FormComponent>
-          <label>{t("pricing")}</label>
           <div
-            className={css`
-              display: grid;
-              grid-template-columns: repeat(3, minmax(0, 1fr));
-              gap: 0.75rem;
-            `}
+            aria-labelledby="label-pricing"
+            className="flex [&_input:focus-visible+label>button]:outline-[5px] [&_input:focus-visible+label>button]:[outline-style:auto] [&_input:focus-visible+label>button]:outline-[Highlight] [&_label:first-of-type>button]:rounded-e-none! [&_label~label>button]:rounded-none! [&_label:last-of-type>button]:rounded-e-(--mi-border-radius)!"
```
timglorioso: I feel like I should change this to normal CSS, but also wanted someone else to witness this Tailwind monstrosity

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PaymentSlider.tsx
```diff
@@ -127,6 +129,7 @@ const PaymentSlider: React.FC<{
             {current >= 15 && t("redistribution")}
           </Pill>
           <ArtistButton
+            aria-label="help"
```
timglorioso: 1. This should be translated
2. Probably more effective: "What is this?"

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PriceAndSuch.tsx
```diff
@@ -225,18 +240,34 @@ const PriceAndSuch: React.FC<{
...
-                  }
-                  disabled={isUpdatingPricingMode}
-                  variant={isSelected ? "default" : "outlined"}
-                >
-                  {pricing.label}
-                </ArtistButton>
+                <React.Fragment key={pricing.key}>
+                  <input
+                    checked={isSelected}
+                    className="sr-only"
```
timglorioso, line 277: TIL Tailwind has a class for visually hidden: `sr-only` (short for screenreader only). Pretty cool!

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PriceAndSuch.tsx
```diff
@@ -225,18 +240,34 @@ const PriceAndSuch: React.FC<{
...
+                    }
+                    ref={pricingModeInputRefs[pricing.key]}
+                    type="radio"
+                  />
+                  <label htmlFor={id}>
+                    <ArtistButton
+                      onClick={() =>
+                        onSelectPricingMode(pricing.key as PricingMode)
+                      }
+                      tabIndex="-1"
```
timglorioso: This is one of those situations where the `input`s are visually hidden and used by keyboards and screenreaders, so we remove the `button`s from the tab order.

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PriceAndSuch.tsx
simonv3: I had a tailwind bug recently that I decided to just write as CSS haha

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/SavingInput.tsx
```diff
@@ -213,19 +213,6 @@ const SavingInput: React.FC<{
...
-          right: 0.5rem;
-          top: 0.5rem;
-          z-index: 99999;
-          -webkit-filter: invert(100%);
-          filter: invert(100%);
-        `}
-      >
-        {isSaving && <LoadingSpinner fill={colors?.primary} size="small" />}
-        {saveSuccess && <FaCheck fill={colors?.primary} />}
-      </div>
```
timglorioso, line 219: To remove the spinner, there needs to be the autosave note on all forms where SavingInput is used. In addition to Release form, it's also used on Profile, Post, Merch forms. I'll be updating those in the next release, so I'll revert and wait to apply this change. In that release I should also reuse the translation string for the note/instruction.

client/src/components/ManageArtist/ManageTrackGroup/ManageTrackGroup.tsx
```diff
@@ -168,6 +169,9 @@ const ManageTrackGroup: React.FC<{}> = () => {
             </ArtistButton>
           </div>
         </SpaceBetweenDiv>
+        <ArtistBox variant="info">
+          <p>{t("instructionFormAutoSave")}</p>
```
timglorioso: I forgot that to do this we need to do something about tracks not autosaving. Either make them autosave or add a note in track editor to save changes.

timglorioso: I think I'm gonna revert this and keep status quo on autosave/save while we work out a new design based on recent feedback.

timglorioso:
@simonv3 wanna give this a peek? I'm having trouble getting dev env working on macOS right now so will test on prod :grimacing:

client/src/components/common/Button.tsx
```diff
@@ -275,7 +275,7 @@ export const Button: React.FC<
       disabled={disabled}
       onlyIcon={!children || onlyIcon}
       title={children ? children.toString() : undefined}
-      aria-label={children ? children.toString() : ""}
+      aria-label={props["aria-label"] || (children ? children.toString() : "")}
```
simonv3, line 278: This is not a note for this PR but I noticed that `children.toString()` doesn't play super nice with if there's a component as a child as a button (example for some reason we put an SVG component in there, which I don't think makes sense, but whatever). Then it'll appear like `[object Object], button text`. Not sure what the fix is here. Could just be that we only let children of buttons be strings.

timglorioso: Thanks for the context. TBH I want to redo that and not use children for title and aria-label automatically.

simonv3 (dismissed):

Issue #1832, opened 2026-03-30
Notify user that their annual subscription will be renewed in a month.


simonv3:
When a user is subscribed on an annual period, we want to give them a heads up that their subscription is due to renew. 

1. After receiving an invoice payment, store the next invoice date on our invoice model.
2. Store a renewalRemidnerSentAt to avoid duplicate renewals being sent.
3. In our every day cron job, create a job that adds a queue worker to send the e-mail
4. Write tests that try to approximate this.

PR #1840, opened 2026-04-02
feat: subscription renewals

### LINKED ISSUE #1830: Digital subscriptions should use the artist-purchase-notification e-mail
Currently these use a custom e-mail. We can instead use the artist-purchase-notification e-mail so that it all uses the same format.

1. Transition handleInvoice to use the artist-purchase-notification e-amil
2. Update the artist-purchase-notification e-mail to display subscription payments.

simonv3:
closes #1830 #1832

### DIFF
--- a/src/jobs/send-subscription-renewal-reminders.ts
+++ b/src/jobs/send-subscription-renewal-reminders.ts
@@ -1,0 +1,142 @@
+import prisma from "@mirlo/prisma";
+import logger from "../logger";
+import { Artist } from "@mirlo/prisma/client";
+import { sendMailQueue } from "../queues/send-mail-queue";
+
+export type SubscriptionRenewalReminderEmailType = {
+  interval: "MONTH" | "YEAR";
+  artist: Artist;
+  host: string;
+  client: string;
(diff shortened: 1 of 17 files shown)


src/jobs/send-subscription-renewal-reminders.ts
```diff
@@ -0,0 +1,148 @@
...
+                  year: "numeric",
+                  month: "long",
+                  day: "numeric",
+                }
+              )
+            : "unknown date";
+
+          // Use transaction to ensure both queue job and timestamp update succeed together
+          await prisma.$transaction(async (tx) => {
+            // Queue the email job
```
simonv3: Should we actually use a queue here?

src/utils/handleFinishedTransactions.ts
```diff
@@ -274,11 +276,16 @@ export const handleTrackGroupPurchase = async (
         },
       });
 
+      console.log("trackGroup", trackGroup.artist);
```
simonv3: Oops remove

PR #1841, opened 2026-04-03
fix: a bunch of style and UX fixes for label and artist pages

### LINKED ISSUE #1838: Clean up album management page for labels
Currently a label's albums shows up both under label albums and artist albums for that label. Need to make the distinction clearer:

Image

simonv3:
closes #1838

There's more to improve but generally tidied things up on the label page and on label and artist release management

* Replaced TrackGroupCard with tables. Though we might want to switch that flex grids that switch to columns on smaller screens. Either way right now it looks tidier.
* Cleaned up logic around what artists get shown where and also showed artists who aren't in the label for easy adding.
* Removed the "manage artists" link from the header menu cause it was confusing and doing the same thing twice.

image
image


client/src/components/ManageArtist/ManageArtistAlbums.tsx
```diff
@@ -1,21 +1,150 @@
...
+        console.error(e);
+        snackbar(t("somethingWentWrong"), { type: "warning" });
+      }
+    },
+    [t]
+  );
+
+  return (
+    <Table className="text-xs">
+      <thead>
```
simonv3: Will switch these away from tables in the morning. No bueno on mobile.

timglorioso: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/HTML_table_basics#when_should_you_avoid_html_tables

simonv3: @timglorioso do you think this qualified as tabular data?

timglorioso: I think a table could be helpful here if you're working with a lot of releases. For example, if I want to ensure that all releases have a catalog number, a table would allow me to scan the rows top-to-bottom and spot any empty cells.

But generally, no, I wouldn't consider something like this tabular data by default. If the table is long and the headers don't stick, then you have to remember which column is which. IMO better to just have a list where each item attribute is contextualized by the proximity to the item name (which can be larger and easier to identify) and by use of descriptive copy. For example, instead of "Release date: 2025-01-09", each list item could say "Released 2025-01-09".

Regardless, if it looks like a table but isn't using `table` then screenreaders might not have as nice of an experience.

timglorioso: Lol just saw Roberta's feedback about the table, glad my catalog number example is real. I think it'd be good to consider bringing back the list view, improving it, and offering it alongside a table view.

simonv3: For what it's worth the table was removed and it uses `grid` https://github.com/funmusicplace/mirlo/pull/1841/changes/ddddc3e60b5a54c3acf959a0bed4f6eb4d979191

timglorioso: I saw that, but for screenreaders we should use `table` if the content is laid out like a table. See https://www.w3.org/WAI/tutorials/tables/#why-is-this-important

Indeed getting `table` to render well on small screens is a challenge.

PR #1842, opened 2026-04-03
fix: simplify welcome screen

simonv3:
This fixes the skip button and the next buttons being disabled on the new artist flow.

### DIFF
--- a/client/src/components/ManageArtist/Welcome.tsx
+++ b/client/src/components/ManageArtist/Welcome.tsx
@@ -27,21 +27,6 @@ interface FormData {
   confirmContentPolicy: boolean;
 }
 
-const nextButtonText = (step: number, currentStepValue?: unknown) => {
-  switch (steps[step]) {
-    case "name":
-      if (currentStepValue) {
-        return "next";
-      }
-      return "enterANameToGetStarted";
-    case "urlSlug":
-      return "customizeYourPage";
-
-    default:
-      return "next";
-  }
-};
-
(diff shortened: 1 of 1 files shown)


client/src/components/ManageArtist/Welcome.tsx
```diff
@@ -78,7 +62,11 @@ const Welcome = () => {
             }
           );
           setLocalArtist(response.result);
-          reset(response.result);
+          reset({
+            name: response.result.name,
+            urlSlug: response.result.urlSlug,
+            confirmContentPolicy: true,
```
simonv3, line 68: This is the main issue, the form was being reset to the confirmContentPolicy being unchecked because it was just inhering from the content on the api response.

timglorioso: Thanks for pointing this out. I guess this got undone at some point, because I just added it back (at least `confirmContentPolicy` alone seemed to fix the issue with name and slug being filled and valid)

Edit: just noticed new PR

client/src/components/ManageArtist/Welcome.tsx
```diff
@@ -148,6 +132,19 @@ const Welcome = () => {
             />
           </FormComponent>
 
+          {step === 0 && (
```
timglorioso, line 135: Breaking these step buttons out feels great, I will include this

timglorioso (approved):

PR #1866, opened 2026-04-09
Feed notifications refactor

LouisLouiseKay:
So this is a big refactor of the UI of the "notification feed". I tried to stick to Tim's guidelines around accessibility, so I did tab through everything, avoided using "states" when unnecessary (since they don't "announce change" and that's bad with readers. Since this is now a "two column layout", I also added a system to skip first column to go to the second based on [this recommendation](https://www.a11y-collective.com/blog/skip-to-main-content/) because otherwise I had to tab to the bottom of hell before reaching the second column. Also based on [these recommendations](https://allyant.com/blog/mastering-the-aria-tabindex-attribute-for-enhanced-web-accessibility/) and good old [html references](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex ) I tried to fix the H1 issue best I could. I think this matches expected behavior from an accessibility standpoint.

The layout itself should be self explanatory, the hardest part was fetching all that new data all around the database to make this expandable for future use cases like AP/Social Web notifications, and handling the pagination for both columns seperately to favor accessibility.


client/src/components/Profile/UserNotificationFeed/LabelInvite.tsx
```diff
@@ -1,66 +1,75 @@
...
-    <div
-      className={css`
-        display: flex;
-        flex-direction: column;
-        p {
-          margin-bottom: 0.5rem;
-          display: inline-block;
-        }
+    <div className="flex items-start gap-3.5 py-3.5 px-4">
+      <div className="w-10 h-10 rounded-full shrink-0 overflow-hidden bg-[#8B8083] flex items-center justify-center font-bold text-sm text-white">
```
simonv3: This should probably use a theme color variable? 
```suggestion
      <div className="w-10 h-10 rounded-full shrink-0 overflow-hidden bg-(--mi-whatever-color) flex items-center justify-center font-bold text-sm text-white">
```

client/src/components/Profile/UserNotificationFeed/LabelInvite.tsx
```diff
@@ -1,66 +1,75 @@
...
-          to={getArtistManageUrl(notification.artist.id) + "/customize#labels"}
-        >
-          {notification.artist?.name}
-        </ButtonLink>
-      </p>
+      <div className="flex-1 min-w-0">
+        <div className="text-xs font-bold uppercase tracking-[0.08em] text-[#8B8083] mb-1">
+          {t("labelTag")}
+        </div>
+        <p className="mb-2 inline-block">
```
simonv3: So a thing that I have learned recently that just generally flows better is to not use padding on child elements but to instead use gaps in the parent div.

client/src/components/TrackGroup/ReleaseDate.tsx
```diff
@@ -27,6 +27,36 @@ export const formatDate = ({
...
 };
 
+export const formatRelativeTime = ({
+  date,
+  i18n,
+}: {
+  date: string;
+  i18n: i18n;
+}): string => {
+  const now = Date.now();
```
simonv3, line 37: this feels like something we should use a library for, might be fine not to.

client/src/components/Profile/UserNotificationFeed/NotificationFeedItem.tsx
```diff
@@ -1,94 +1,69 @@
...
-
+import FundraiserPledge from "./FundraiserPledge";
 import LabelInvite from "./LabelInvite";
-import { set } from "lodash";
 
-const LI = styled.li<{ isRead: boolean }>`
+const categoryBorderColor = (type: Notification["notificationType"]) => {
+  switch (type) {
+    case "NEW_ARTIST_ALBUM":
+      return "#BE3455"; // Red 500
```
simonv3, line 14: Should these be theme colors?

client/src/components/Profile/UserNotificationFeed/UserBoughtYourAlbum.tsx
```diff
@@ -1,49 +1,110 @@
...
-    </>
+        >
+          <strong>{buyerName}</strong> {t("bought")}{" "}
+          <Link to={releaseUrl}>
+            <strong>{notification.trackGroup.title}</strong>
+          </Link>
+        </div>
+        {notification.trackGroup.artist && (
+          <div className="text-xs text-(--mi-light-foreground-color) mt-0.5">
+            <span>{t("by")} </span>
```
simonv3, line 80: Not urgent but _ideally_ this uses the translation component <Trans> because in other languages the order of the text might be different.

client/src/components/Profile/UserNotificationFeed/UserBoughtYourAlbum.tsx
```diff
@@ -1,49 +1,110 @@
...
+            <span
+              className={
+                compact
+                  ? "inline-block text-xs font-semibold py-0 px-1.5 rounded-full bg-[#E1F4FC] text-[#34515C] mt-0.5 whitespace-nowrap"
+                  : "inline-block text-xs font-semibold py-0.5 px-2 rounded-full bg-[#E1F4FC] text-[#34515C] mt-1 whitespace-nowrap"
+              }
+            >
+              {(
+                notification.trackGroup.purchase.transaction.amount / 100
+              ).toLocaleString(undefined, {
```
simonv3: use the `moneyDisplay` function or component `<Money/>` that handles this.

client/src/queries/notifications.ts
```diff
@@ -1,6 +1,9 @@
 import { QueryFunction, queryOptions } from "@tanstack/react-query";
 import * as api from "./fetch/fetchWrapper";
-import { QUERY_KEY_NOTIFICATIONS, QUERY_KEY_TAGS } from "./queryKeys";
+import { QUERY_KEY_NOTIFICATIONS } from "./queryKeys";
+
+export const markAllNotificationsRead = (userId: number) =>
+  api.put(`v1/users/${userId}/notifications/markAllRead`, {});
```
simonv3: Do you have prettier installed and running on save? This feels like a weird formatting to me.

client/src/translation/en.json
```diff
@@ -1147,13 +1147,39 @@
...
+    "purchaseTag": "Purchase",
+    "labelTag": "Label",
+    "startedFollowingYou": "started following you",
+    "subscribedToYou": "subscribed to you",
+    "bought": "bought",
+    "by": "by",
+    "goToRelease": "Go to release",
+    "readPost": "Read post",
+    "earlierNotifications": "Last read",
+    "timelineFollowing": "Artists you follow",
```
simonv3, line 1170: How do we make this not "artist" specific? IN other places we've removed "Artist" language to be more general.

src/routers/v1/users/{userId}/notifications/markAllRead.ts
```diff
@@ -0,0 +1,50 @@
...
+export default function () {
+  const operations = {
+    PUT: [userAuthenticated, PUT],
+  };
+
+  async function PUT(req: Request, res: Response) {
+    const { userId } = req.params as unknown as Params;
+    const loggedInUser = req.user as User;
+
+    if (Number(userId) !== Number(loggedInUser.id)) {
```
simonv3: There's middleware for this that we should use more `userHasPermission("owner")`.

It's an issue in all the `/user/{userId}/endpoints` tbh.

client/src/components/Profile/UserNotificationFeed/LabelInvite.tsx
LouisLouiseKay: Ah yeah, wanted to touch on that : I think our current colors are a bit limiting in many context (especially when we want to contrast more and use colors for that). What I did was use the "Color palette" that we've devised in our "Mirlo Styling" document. Obviously I feel like the "better option" would be to just add more colors to the Theme color variables.

client/src/components/Profile/UserNotificationFeed/LabelInvite.tsx
LouisLouiseKay: Ah ! Getting reacquainted with Tailwind and really not used yet to best practices. OK will fix

client/src/components/Profile/UserNotificationFeed/UserBoughtYourAlbum.tsx
LouisLouiseKay: Oh I didn't even think of that, I just was smart enough to remember to seperate the link from the translatable string because it caused issue once.

client/src/translation/en.json
LouisLouiseKay: I feel like I'd like there to remain a distinction between "stuff I follow because it's the makers of art pieces I voluntarily followed and I want to check" and "stuff I follow that's curation of artists I might want to check"

client/src/queries/notifications.ts
LouisLouiseKay: I have and I did save with prettier installed... normally? What's odd about the formatting? Wouldn't be the first time prettier doesn't do what it's supposed to on my local install.

simonv3: Ideally it breaks that onto multiple lines IMO

client/src/components/Profile/UserNotificationFeed/LabelInvite.tsx
simonv3: Right, I think you should add them!

client/src/components/Profile/UserNotificationFeed/index.tsx
```diff
@@ -1,54 +1,296 @@
...
+          >
+            {opt.label}
+          </span>
+        </label>
+      ))}
+    </div>
+  </fieldset>
+);
+
+const NotificationColumn: React.FC<{
```
simonv3: Can these be broken into their own files?

client/src/styles/index.css
```diff
@@ -9,6 +9,10 @@ body {
   margin: 0;
 }
 
+h1[tabindex="-1"]:focus {
+  outline: none;
```
timglorioso: This is already taken care of by lines 78-80. Instead of adding a duplicate rule, I recommend instead removing 83-86 (the focus-visible rules).

client/src/components/Profile/UserNotificationFeed/index.tsx
```diff
@@ -1,54 +1,296 @@
...
+  const { t } = useTranslation("translation", { keyPrefix: "notifications" });
+  const [filter, setFilter] = React.useState("all");
+  const { page, resetPage, PaginationComponent } = usePagination({
+    pageSize,
+    pageParam: filterName,
+  });
+
+  // Conditional options need individual existence checks (can't call hooks in a loop)
+  const cOpts = filterOptions.filter((o) => o.checkExists);
+  const cOpt0 = cOpts[0];
```
simonv3: Why are we filtering by options that have this boolean and then specifically calling them? Why not just have an array that uses them directly? It's taken me much too long to try and understand how this is working (which I don't like for if I have to debug it) or what it's trying to do. 

It feels like there needs to be a query at a higher level to determine if there are notifications of a type so that a filter doesn't display everything relevant? I think what's making it complicated to follow is that we're using these slightly meaningless cOpt0 but we know there's only two of them.

client/src/styles/index.css
LouisLouiseKay: OK Yeah no problem !

client/src/components/Profile/UserNotificationFeed/index.tsx
LouisLouiseKay: Yeah ok, and actually we should do this for all categories of notifications because ultimately it was only because I thought of "labels" and "pledges" as being very case specific, but all items are empty when you create your profile for the first time, and I don't want to populate that area with empty labels that link to nothing (which was the point initially).

client/src/components/Profile/UserNotificationFeed/index.tsx
LouisLouiseKay: Yeah, absolutely, I'll do that.

src/routers/v1/users/{userId}/notifications/markAllRead.ts
LouisLouiseKay: OK I had no idea ! normally this is fixed in the next commit.

client/src/components/Profile/UserNotificationFeed/index.tsx
```diff
@@ -1,54 +1,173 @@
...
-  const { page, PaginationComponent } = usePagination({ pageSize });
-  const {
-    data: notifications,
-    refetch,
-    isPending,
-  } = useQuery(
-    queryNotifications(user?.id, { skip: pageSize * page, take: pageSize })
-  );
+
+  const hasNotifications = (data: { total?: number } | undefined): boolean =>
```
simonv3: Not sure this needs to be a function but if it is could definite it outside of the component (just above it) to keep the code here a little cleaner.

client/src/components/Profile/UserNotificationFeed/index.tsx
```diff
@@ -1,54 +1,173 @@
...
+            types: ["LABEL_ADDED_ARTIST"] as Notification["notificationType"][],
+          },
+        ]
+      : []),
+  ];
+
+  React.useEffect(() => {
+    if (!user?.id) return;
+    return () => {
+      markAllNotificationsRead(user.id);
```
simonv3, line 116: I like it.

simonv3 (approved):
lgtm!

PR #1879, opened 2026-04-11
fix: subscribe box visual fixes

LouisLouiseKay:
It was using "darken color" for the outline, so the borders were invisible on dark backgrounds. Now it uses theme's primary and secondary as its base. I also added a slight overlay background color on the sides of the section to both make the "arrow" more visible and clarify carousel navigation.

### DIFF
--- a/client/src/components/Artist/ArtistSupportBox.tsx
+++ b/client/src/components/Artist/ArtistSupportBox.tsx
@@ -1,14 +1,11 @@
 import { css } from "@emotion/css";
 import React from "react";
 import api from "services/api";
-import Box from "../common/Box";
 import Money from "../common/Money";
 import { useTranslation } from "react-i18next";
 import MarkdownContent from "components/common/MarkdownContent";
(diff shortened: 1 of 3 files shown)


simonv3:
What witchcraft are these images lol

client/src/components/Artist/ArtistSupportBox.tsx
```diff
@@ -82,37 +79,38 @@ const ArtistSupportBox: React.FC<{
...
-          </div>
-        )}
+    <div
+      className={
+        (isSubscribedToTier ? " border-4" : " border-1") +
+        " relative border-inset pb-5 mb-1 gap-5 bg-(--mi-darken-background-color) flex flex-col text-sm border-(--tier-border-color)"
+      }
+      style={
+        {
+          "--tier-border-color": tierBorderColor,
```
simonv3, line 100: ah, clever?

simonv3 (approved):

PR #1883, opened 2026-04-12
fix: allow label releases to be linked to merch items

peterklingelhofer:
The API validation was rejecting label releases because it only checked artistId

Now also accepts track groups where the logged-in user is the label owner

Closes #1859

### DIFF
--- a/src/routers/v1/manage/merch/{merchId}/index.ts
+++ b/src/routers/v1/manage/merch/{merchId}/index.ts
@@ -81,12 +81,17 @@ export default function () {
 
       if (newValues.includePurchaseTrackGroupId) {
         // check that the artist who owns this merch also
-        // owns this trackgroup
-
+        // owns this trackgroup, or that the logged-in user
+        // is the label owner of the trackgroup
+        const user = req.user as User;
         const trackGroup = await prisma.trackGroup.findFirst({
           where: {
-            artistId: merch?.artistId,
             id: Number(newValues.includePurchaseTrackGroupId),
+            OR: [
+              { artistId: merch?.artistId },
+              { paymentToUserId: user.id },
+              { artist: { userId: user.id } },
+            ],
           },
         });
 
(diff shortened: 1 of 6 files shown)


src/routers/v1/manage/merch/{merchId}/index.ts
```diff
@@ -81,12 +81,17 @@ export default function () {
 
       if (newValues.includePurchaseTrackGroupId) {
         // check that the artist who owns this merch also
-        // owns this trackgroup
-
+        // owns this trackgroup, or that the logged-in user
+        // is the label owner of the trackgroup
+        const user = req.user as User;
```
peterklingelhofer, line 87: Noticed this pattern is replicated 110 times in the codebase - leveraging the Express global here helps us always retain parity with the Passport type easily. Didn't do the find and replace in this MR because I didn't want merge conflicts, can do it as a quick and clean one-off refactor cleanup PR later

src/routers/v1/manage/merch/{merchId}/index.ts
```diff
@@ -81,12 +81,17 @@ export default function () {
...
         const trackGroup = await prisma.trackGroup.findFirst({
           where: {
-            artistId: merch?.artistId,
             id: Number(newValues.includePurchaseTrackGroupId),
+            OR: [
+              { artistId: merch?.artistId },
+              { paymentToUserId: user.id },
+              { artist: { userId: user.id } },
```
simonv3: An additional case here, and its one that only applies if the current artist is a label, but it's if the merch belongs to an artist that's a `isLabelProfile`, we should also be looking in that artist's user's artistLabels.

And I'm wondering if there's a way to re-use the logic of `artistBelongsToLoggedInUser` here. Or to standardize some function that's like `whereForAllArtistsThisLabelCanEdit`

simonv3:
Love the type update, lmk what you think about extra query ask

simonv3 (approved):

Issue #1333, opened 2025-07-15
On artist profile, add a way to apply platform percent to all existing releases


simonv3:
As it says on the tin. 

Otherwise an artist has to go through all releases and change it if they want to update it.

timglorioso:
Smells like modeling. Typically platform cut is something you'll want the same for most releases, with some exceptions. The desire to change all of them to the same value is evidence of this.

I suggest storing platform cut per artist, then allow overriding per release. TBD how we'd migrate existing data, anticipating something to do there.

simonv3:
Yeah, artists currently have a `defaultPlatformFee` and now when an artist adds an album, it sets it to that number _for that album_ specifically. But the ask here is that someone might want to change the number for all existing albums. So I think the original poster asked something like "add a toggle that will set my platform fee for all albums if I change it in the default".

So no need to migrate I think, just let artists choose if they want to bulk change platform fees for their albums, or just for future ones.

simonv3:
Because someone might want to have a specific trackgroup be more or less than the average, or on a specific day for it to be different. I think if it overwrites existing `trackGroupPlatformFees` by default (or rather, as a non-option) that would be confusing.

timglorioso:
Here's what I'm imagining:

1. Artist sets platform fee while creating first release. We silently save this as the default for all this artist's releases.
2. Any future releases use that default, or can be set to a custom fee.
3. If a release has a custom fee set, it is not affected by changes to the default.
4. If a release does not have a custom fee set, it is affected by changes to the default.

timglorioso:
Coming back to this, I hope it's clear what I'm thinking. Changing the default should change it for all releases except those where the default is overwritten by a custom fee. You shouldn't need to choose which releases get the new default value. Custom fee on a release indicates you don't want it to get the new default.

This approach sidesteps issues like missing selecting some releases when changing the default. It takes care of updating all past and future releases using the default.

We might present the list of releases that will be updated with links so the user can set any custom fees before proceeding with the bulk action.

PR #1891, opened 2026-04-14
feat: add defaultIsPreview to TrackGroup for bulk track preview toggling

peterklingelhofer:
Closes #1857


client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ManageTrackDefaults.tsx
```diff
@@ -57,6 +77,26 @@ const ManageTrackDefaults: React.FC<BulkUpdateTracksProps> = ({
           </label>
         </ToggleFormComponent>
       </FormProvider>
+      {(trackGroup.tracks ?? []).length > 0 && (
+        <div
+          className={css`
```
simonv3: Can you use tailwind instead of these?

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ManageTrackDefaults.tsx
```diff
@@ -25,9 +26,28 @@ const ToggleFormComponent = styled("div")`
...
+  reload,
 }) => {
   const { t } = useTranslation("translation", { keyPrefix: "manageAlbum" });
   const methods = useForm<TrackGroup>({ defaultValues: trackGroup });
+  const snackbar = useSnackbar();
+
+  const handleSetAllTracksPreview = async (isPreview: boolean) => {
+    try {
+      await Promise.all(
+        (trackGroup.tracks ?? []).map((track) =>
```
simonv3: Do we want to overwrite existing settings for tracks? I'm not sure I have a good idea and we could do this for now and see whether we get feedback.

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ManageTrackDefaults.tsx
```diff
@@ -57,6 +77,26 @@ const ManageTrackDefaults: React.FC<BulkUpdateTracksProps> = ({
...
+      {(trackGroup.tracks ?? []).length > 0 && (
+        <div
+          className={css`
+            display: flex;
+            gap: 1rem;
+            flex-wrap: wrap;
+            padding: 0.5rem 0;
+          `}
+        >
+          <ArtistButton onClick={() => handleSetAllTracksPreview(true)}>
```
simonv3: A bit tied in to the above comment, I'm not sure about the UX expectation here.

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ManageTrackDefaults.tsx
peterklingelhofer: Hmmm, I thought that was the goal - if the user uses this option they want to set all of these tracks to a certain setting, regardless of what they were previously set to. If I'm misunderstanding though please elaborate

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ManageTrackDefaults.tsx
peterklingelhofer: Fixed in https://github.com/funmusicplace/mirlo/pull/1891/commits/8751f1625d21954d7bf4ad6108b09c6cc59269a1

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ManageTrackDefaults.tsx
simonv3: The initial idea was that this would set the default for future uploads (because this happens earlier in the form, tracks are only uploaded at the end). I think a user would be surprised if this overwrote a setting they had already set?

simonv3 (dismissed):

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ManageTrackDefaults.tsx
peterklingelhofer: Something like this? https://github.com/funmusicplace/mirlo/pull/1891/changes/178f171dda56a464773802a18eed3037408f814e

timglorioso: I posted some thoughts to this related issue: https://github.com/funmusicplace/mirlo/issues/1333

peterklingelhofer: Backed out the last commit per discussion

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ManageTrackDefaults.tsx
peterklingelhofer: As above, is this discussion now resolved?

simonv3 (approved):

PR #1900, opened 2026-04-15
fix: allow testOwns to return true for preview tracks purchased via a…

### LINKED ISSUE #1892: Allow previewed tracks to be downloaded if the user has pre-ordered an album
When a user has bought an album in pre-order (before release date, after publication date) they should be able to download tracks that are listed as `isPreview`.

peterklingelhofer:
…lbum

Closes #1892

### DIFF
--- a/client/cypress/e2e/preview-track-downloads.cy.ts
+++ b/client/cypress/e2e/preview-track-downloads.cy.ts
@@ -18,6 +18,8 @@ const pastReleaseDate = new Date(
 ).toISOString();
 
 describe("preview track downloads with album purchase", () => {
+  let customerId: number;
+  let artistId: number;
   let preOrderTrackGroupId: number;
   let previewTrackId: number;
   let nonPreviewTrackId: number;
@@ -26,9 +28,6 @@ describe("preview track downloads with album purchase", () => {
   before(() => {
     cy.task("clearTables");
 
-    let customerId: number;
-    let artistId: number;
-
     cy.task("createUser", {
       email: artistOwnerEmail,
       password: artistOwnerPassword,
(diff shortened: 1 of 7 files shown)


simonv3:
I'm looking at the code and I'm not actually sure this gets used atm. Looking at the TrackView page it looks like downloads is done in `<PurchaseOrDownloadAlbum>` with checking against the `user.userTrackPurchases` and that's just a select based on if the user bought the track. There isn't actually a check for whether the track is published or released, which I think is a problem (should definitely check if it's published). So it feels like there's currently a bug here. 

You can download a track as a pre-order (desired, but an accident).
You can download an album as a pre-order (not desired, also an accident, it shows up in your purchases). 

This feels like the exact scenario for an cypress e2e test to make sure this will work going forward.

simonv3 (approved):

client/cypress/e2e/preview-track-downloads.cy.ts
```diff
@@ -0,0 +1,160 @@
...
+
+        return cy.task("createTrack", {
+          title: "Preview Track",
+          urlSlug: "preview-track",
+          trackGroupId: tg.id,
+          isPreview: true,
+          allowIndividualSale: true,
+        });
+      })
+      .then((track) => {
```
simonv3: This is so verbose haha

simonv3 (dismissed):
Really appreciate the CY test. Good sanity shit.

client/cypress/e2e/preview-track-downloads.cy.ts
peterklingelhofer: Yeah let me clean this up before we merge, this should be await

simonv3: Oh, I think that's a CY thing not necessarily a you thing, but if you can get it working cool!

simonv3 (approved):

client/cypress/e2e/preview-track-downloads.cy.ts
peterklingelhofer: Ah you're right, Cypress docs state that await is unreliable here so I'll leave it

PR #1901, opened 2026-04-15
Refactor renaming banner into background

LouisLouiseKay:
OK so this was long overdue (and I really should have done it since the beginning when repurposing the banner into an actual background), but with the all the recent talks of needing to offer more styling options and allow the page to feel more like a proper website, it felt extremely important to rename this correctly because it's currently misleading in the database. @simonv3 definitely check that I've done it exactly as it should be, normally only the minIO/S3 "bucket names" ("incoming-artist-banners" and "artist-banners") are left with the incorrect naming and that's the only part you'll have to do by yourself on production. Otherwise it should just work as is but I'll let you review this, it was a pretty significant change cause there were a lot of mention of it kind of everywhere.

### DIFF
--- a/client/src/components/Artist/ArtistLinks.tsx
+++ b/client/src/components/Artist/ArtistLinks.tsx
@@ -37,7 +37,7 @@ const ArtistLinks: React.FC = () => {
   const isOwningArtist = artist?.userId === user?.id;
 
   return (
-    // <ArtistPageWrapper artistBanner={!!artistBanner}>
+    // <ArtistPageWrapper hasBackground={!!hasBackground}>
     <div
       className={css`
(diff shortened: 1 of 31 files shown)


client/src/components/Artist/ArtistLinks.tsx
```diff
@@ -37,7 +37,7 @@ const ArtistLinks: React.FC = () => {
   const isOwningArtist = artist?.userId === user?.id;
 
   return (
-    // <ArtistPageWrapper artistBanner={!!artistBanner}>
+    // <ArtistPageWrapper hasBackground={!!hasBackground}>
```
simonv3, line 40: Since this is a comment just delete the line?

PR #1904, opened 2026-04-15
Refactor renaming banner into background

LouisLouiseKay:
With the all the recent talks of needing to offer more styling options and allow the page to feel more like a proper website, it felt extremely important to rename background correctly because its naming was currently misleading in the database. Only the minIO/S3 "bucket names" ("incoming-artist-banners" and "artist-banners") are left with the incorrect naming because they'll eventually be abandonned for another system anyway.

### DIFF
--- a/client/src/components/ManageArtist/UploadGeneralImage.tsx
+++ b/client/src/components/ManageArtist/UploadGeneralImage.tsx
@@ -17,63 +17,6 @@ import { ArtistButton } from "components/Artist/ArtistButtons";
 import { useAuthContext } from "state/AuthContext";
 import { useFormContext } from "react-hook-form";
 
-type ImageType = "banner" | "avatar" | "cover" | "image" | "profile";
-
-export function isTrackgroup(entity: unknown): entity is TrackGroup {
-  if (!entity) {
-    return false;
-  }
-  return (entity as TrackGroup).cover !== undefined;
-}
-
-export function isUser(entity: unknown): entity is LoggedInUser {
-  if (!entity) {
-    return false;
-  }
-  return (entity as LoggedInUser).userAvatar !== undefined;
-}
-
-export function isTrack(entity: unknown): entity is Track {
-  if (!entity) {
-    return false;
-  }
-  return (entity as Track).trackGroupId !== undefined;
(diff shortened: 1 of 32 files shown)


client/src/components/ManageArtist/UploadGeneralImage.tsx
```diff
@@ -62,8 +62,8 @@ const getExistingImage = (
     } else if (isUser(existing) && imageType === "banner") {
       image = existing.userBanner;
     }
-  } else if (imageType === "avatar" || imageType === "banner") {
-    image = existing[imageType];
+  } else if (imageType === "avatar") {
+    image = (existing as Artist).avatar;
```
simonv3: We don't need a case for background here?

client/src/components/ManageArtist/SubscriptionForm.tsx
```diff
@@ -157,7 +157,7 @@ const SubscriptionForm: React.FC<{
           image={existing?.images?.[0]?.image}
           width="100%"
           size={625}
-          dimensions="banner"
+          dimensions="background"
```
simonv3: this is about a subscription

client/src/components/ManageArtist/UploadGeneralImage.tsx
LouisLouiseKay: Normally no this is for the user banner stuff... which ultimately we should remove but I think it's out of scope.

client/src/components/common/UserBanner.tsx
```diff
@@ -16,14 +16,14 @@ const UserBanner = () => {
 
   const { data: label } = useQuery(queryLabelBySlug(labelSlug));
 
-  const banner = label?.banner;
+  const banner = label?.background;
```
simonv3: I still think this shouldn't be changed.

client/src/components/ManageArtist/UploadGeneralImage.tsx
simonv3: But I don't see a handler for the background now?

simonv3 (approved):

PR #1915, opened 2026-04-16
fix: remove currency dropdown, inherit payout currency from Stripe

### LINKED ISSUE #1910: Dropdown on managing currencies
> When I first set up the account I set the currency as CAD. But later Ian informed me all prices should be in USD. After completing the Stripe set up I saw an option to set the currency, so I changed it to USD. However, now I'm setting up the Subscription tier and the price is still showing "in cad".
> 
> I searched throughout the interface to see where I set the currency to confirm it was still set to USD, maybe I forgot to hit save(?), but I can't for the life of me figure out where to confirm/change the currency setting. ha!

I _think_ we should just remove the currency dropdown altogether. It will always be managed by Stripe, and we should inherit from Stripe after completion of Stripe set up flow.

peterklingelhofer:
Closes #1910

### DIFF
--- a/client/src/components/ManageArtist/Manage.tsx
+++ b/client/src/components/ManageArtist/Manage.tsx
@@ -5,7 +5,6 @@ import api from "services/api";
 import { ButtonLink } from "../common/Button";
 import { bp } from "../../constants";
 import { useTranslation } from "react-i18next";
-import CountrySelect from "./CountrySelectForm";
 import WidthContainer from "components/common/WidthContainer";
 import { useAuthContext } from "state/AuthContext";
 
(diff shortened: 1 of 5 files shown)


client/src/components/ManageArtist/Manage.tsx
```diff
@@ -5,7 +5,6 @@ import api from "services/api";
 import { ButtonLink } from "../common/Button";
 import { bp } from "../../constants";
 import { useTranslation } from "react-i18next";
-import CountrySelect from "./CountrySelectForm";
```
simonv3, line 8: Can you actually move this to the AccountContainer?

simonv3: Oh I see you did this in another PR

simonv3 (approved):

peterklingelhofer:
Consolidating into #1924 — that PR now includes both the CurrencySelect move to AccountContainer and the Stripe payout currency display in StripeStatusAndButton.

PR #1926, opened 2026-04-19
perf: bulk PUT /manage/trackGroups/:id/tracks endpoint replaces N per…

peterklingelhofer:
…-track calls

### DIFF
--- a/client/src/utils/tracks.ts
+++ b/client/src/utils/tracks.ts
@@ -13,6 +13,16 @@ export const determineNewTrackOrder = produce(
   }
 );
 
+export const bulkSetTracksIsPreview = async (
+  trackGroupId: number,
+  isPreview: boolean
+) => {
+  return api.put<{ isPreview: boolean }, { result: TrackGroup }>(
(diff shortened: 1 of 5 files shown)


client/src/utils/tracks.ts
```diff
@@ -13,6 +13,16 @@ export const determineNewTrackOrder = produce(
   }
 );
 
+export const bulkSetTracksIsPreview = async (
+  trackGroupId: number,
+  isPreview: boolean
+) => {
```
simonv3: Feels a little redundant to put this in a wrapper.

simonv3 (dismissed):
Looks good just one comment.

client/src/utils/tracks.ts
peterklingelhofer: Something like this: https://github.com/funmusicplace/mirlo/pull/1926/commits/4b76c6b9359aa3171bd2bc9a71acf50cf1b1ea4a?

simonv3 (approved):

PR #1927, opened 2026-04-19
bugfix: Vite plugin PWA

### LINKED ISSUE #1916: Use Vite Plugin PWA to prevent old files from breaking the app
We periodically try loading old files when a user is in the browser already and we make a deploy. If those files don't exist anymore the site crashes. 

We should use Vite Plugin PWA to prevent that https://www.npmjs.com/package/vite-plugin-pwa

https://vite-pwa-org.netlify.app/guide/prompt-for-update.html

peterklingelhofer:
Closes #1916

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -120,6 +120,7 @@
     "terser": "^5.30.3",
     "typescript": "5.8.2",
     "vite": "^5.2.9",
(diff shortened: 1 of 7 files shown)


simonv3 (approved):
Really curious how aggressive this will feel to users.

PR #1931, opened 2026-04-19
feat: persist track uploads across navigation via upload context

### LINKED ISSUE #1909: Warn user that tracks are still uploading if they try to navigate away from the page
> -No warning if you leave a page without the tracks completing upload.

Can we have this be a process that happens outside of the album form? Is there an uploader that tracks while the user is navigating around Mirlo? Seems like a cool feature to implement. Kind of how google handles its uploads. 

Would still need a way to warn on navigating if they literally try to change the URL though.

peterklingelhofer:
Closes #1909

### DIFF
--- a/client/src/components/common/UploadProgressPanel.tsx
+++ b/client/src/components/common/UploadProgressPanel.tsx
@@ -1,0 +1,94 @@
+import React from "react";
+import { css } from "@emotion/css";
+import { useTranslation } from "react-i18next";
+import { FaChevronDown, FaChevronUp } from "react-icons/fa";
+import { useUpload } from "state/UploadContext";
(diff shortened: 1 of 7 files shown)


client/src/components/common/UploadProgressPanel.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  if (!isActive) return null;
+
+  const activeCount = queue.length;
+
+  return (
+    <div
+      role="status"
+      aria-live="polite"
+      className={css`
```
simonv3: Can you make these tailwind

peterklingelhofer: Addressed and fixed merge conflicts

simonv3 (approved):
Looks good to me. Wondering if we still need the old rows showing uploading but I guess for now it's fine. Another future improvement is to link to the album when uploading is done.

PR #1933, opened 2026-04-19
feat: pre-commit formatting linting

### LINKED ISSUE #1823: Minimize pull request noise by ensuring code formatting settings get used
Personally not familiar with linting tools and similar, but IMO the ideal is to run prescribed formatting settings pre-commit.

peterklingelhofer:
Will be many conflicts as this import sorts most files. May need to hold off til after April milestone.

Closes #1823

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,9 @@ client/build
 package-lock.json
 yarn-error.log
 .DS_Store
-.vscode
+.vscode/*
+!.vscode/extensions.json
+!.vscode/settings.json
 .pnp.*
 .yarn/*
 __generated__/
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,0 +1,1 @@
+npx --no -- commitlint --edit "$1"
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,1 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-# Runs prettier on the files specified in package.json / "lint-staged"
 yarn lint-staged
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,6 @@
+{
+  "plugins": ["@trivago/prettier-plugin-sort-imports"],
+  "importOrder": ["^@mirlo/(.*)$", "^[./]"],
+  "importOrderSeparation": true,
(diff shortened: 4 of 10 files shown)


timglorioso:
Thanks @peterklingelhofer for opening this. What would be the workflow to run these checks/formats? Is it tied to git operations? Just as an example, I've been using Vim without any lint/format integration. Would it still be up to me to make sure my editor is doing something, or otherwise run some commands manually?

peterklingelhofer:
> Thanks @peterklingelhofer for opening this. What would be the workflow to run these checks/formats? Is it tied to git operations? Just as an example, I've been using Vim without any lint/format integration. Would it still be up to me to make sure my editor is doing something, or otherwise run some commands manually?

Good question. There are two aspects at play here:

Automatic (tied to git): A Husky pre-commit hook runs lint-staged, which runs `eslint --fix` + `prettier --write` on whatever you staged — so on git commit your changes get formatted/linted before they land, regardless of editor. There's also a commit-msg hook that runs commitlint against the Conventional Commits spec. No manual action needed at commit time.

Editor (optional): The .vscode/settings.json we added turns on format-on-save + ESLint fix-on-save for VS Code users who install the recommended extensions (.vscode/extensions.json prompts them). For Vim (or any editor without integration), you have two options: rely entirely on the pre-commit hook (no setup), or run `yarn lint:fix` / `yarn prettier --write <paths>` manually when you want immediate feedback. Either works, the hook is the backstop that guarantees what lands in git is formatted.

timglorioso:
That sounds amazing.

> There's also a commit-msg hook that runs commitlint against the Conventional Commits spec.

Cool, I hadn't heard of that before https://www.conventionalcommits.org/en/v1.0.0/.

I know it's configurable, but I'm glad to see the default is 100 max length header, those extra 28 characters are crucial :D
https://github.com/conventional-changelog/commitlint/blob/75eca7ccf84c1a76fd435e7930237c0c748b8067/%40commitlint/config-conventional/src/index.ts#L18

I'll try this out this week and see how it goes.

simonv3:
@peterklingelhofer instead of running the linter on all files in the repo, could you redo this PR with just the changes that introduce the linter? And then future PRs can do it bit by bit?

simonv3 (changes requested):
required comment

simonv3 (approved):

PR #1935, opened 2026-04-20
feat: Artist contact form

### LINKED ISSUE #515: Feature request: add city and contact form to artist bio
I'd like to see some more information on artist profiles, namely:

- a contact form for booking requests, etc (this could even just be a link to an email address if we don't want to actually handle sending mail)
- a well-structured city field so we can browse artists by location (keeping in mind there are many cities with the same name across the world, so it shouldn't just be a text field)
    - Maybe something like https://osmnames.org

peterklingelhofer:
Closes #515


src/routers/v1/artists/{id}/contact.ts
```diff
@@ -0,0 +1,146 @@
...
+          description: "This artist is not accepting direct messages",
+        });
+      }
+
+      const recentCount = await prisma.notification.count({
+        where: {
+          notificationType: "ARTIST_CONTACT_MESSAGE",
+          artistId: artist.id,
+          relatedUserId: sender.id,
+          createdAt: { gte: new Date(Date.now() - CONTACT_RATE_WINDOW_MS) },
```
simonv3, line 75: 24 messages in 24 hours actually feels like a lot.

src/routers/v1/artists/{id}/contact.ts
```diff
@@ -0,0 +1,146 @@
...
+import { userAuthenticated } from "../../../../auth/passport";
+import prisma from "@mirlo/prisma";
+import sendMail from "../../../../jobs/send-mail";
+import { Job } from "bullmq";
+import { getClient } from "../../../../activityPub/utils";
+import { AppError } from "../../../../utils/error";
+
+type Params = { id: string };
+
+const CONTACT_RATE_LIMIT = 24;
```
peterklingelhofer:
```suggestion
const CONTACT_RATE_LIMIT = 2;
```

prisma/schema.prisma
```diff
@@ -313,6 +314,7 @@ model Artist {
   paymentToUser                      User?                                @relation(fields: [paymentToUserId], references: [id], name: "paymentToUser")
   isLabelProfile                     Boolean                              @default(false)
   artistLocationTags                 ArtistLocationTag[]
+  allowDirectMessages                Boolean                              @default(true)
```
simonv3, line 317: Default to false I think

client/src/components/common/ArtistHeaderSection.tsx
```diff
@@ -234,7 +235,12 @@ const ArtistHeaderSection: React.FC<{
                     )}
                   </div>
                   <ArtistActions>
```
simonv3, line 237: Maybe add a flex with gap-1 tailwind here so this doesn't look like there's no gap 

Image

simonv3 (commented):
Okay, got a chance to test it out, just two comments.

client/src/components/common/ArtistHeaderSection.tsx
peterklingelhofer: What's the deal with the Tailwind suggestions, does this codebase use Emotion and Tailwind? In the process of migrating to Tailwind?

simonv3 (dismissed):

client/src/components/common/ArtistHeaderSection.tsx
simonv3: In the process of migrating to tailwind.

simonv3 (approved):

PR #1936, opened 2026-04-20
fix: Auto purchase albums release vs publish mixup, test

### LINKED ISSUE #1598: autoPurchaseAlbums not working
We should write a test for this

peterklingelhofer:
Closes #1598

### DIFF
--- a/src/jobs/trigger-auto-purchase-new-albums.ts
+++ b/src/jobs/trigger-auto-purchase-new-albums.ts
@@ -3,22 +3,22 @@ import logger from "../logger";
 import { autoPurchaseNewAlbumsQueue } from "../queues/auto-purchase-new-albums-queue";
 
 /**
- * Trigger function: Finds recent albums and enqueues auto-purchase jobs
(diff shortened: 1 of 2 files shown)


src/jobs/trigger-auto-purchase-new-albums.ts
```diff
@@ -3,22 +3,22 @@ import logger from "../logger";
...
  */
 export async function triggerAutoPurchaseNewAlbums() {
   const currentDate = new Date();
-  const oneHourAgo = new Date();
-  oneHourAgo.setHours(currentDate.getHours() - 1);
+  const oneHourAgo = new Date(currentDate.getTime() - 60 * 60 * 1000);
 
   const recentAlbums = await prisma.trackGroup.findMany({
     where: {
-      releaseDate: {
```
simonv3, line 17: I _thiiiink_ this is going to be reworked in @LouisLouiseKay album flow rejig, and I also think we still only want to check for release dates in the past?

simonv3 (approved):

Issue #1245, opened 2025-06-16
Word choices for `Can be played without purchase ?` column while managing release tracks


Cwpute:
When you're editing a track, in the column `Can be played without purchase ?` you can set "Enable free listen", but then when you're not editing the track it shows instead "Preview". I believe this is because they don't use the same key ? Ideally both should show the same thing so using the same key for both should be fine.

It's also strange that the column is shaped as a question but the value it can be given doesn't answer that question directly !
If the title is closed question like now which can only be answered with slight variations of Yes and No ("Yes, users can listen without purchasing the track.", "No, they can't" or "Yes, they can but..." etc.) then the value/answer should include that clear answer.

My solution would be to give that column the simple name of `Streaming` or more precisely `Online Streaming`, without a question mark, to enable more complex answers: "Unlimited preview plays", "Purchase only", and later on if we build these features "Disabled", "Limited preview plays (3)" etc...

timglorioso:
I had a similar thought as I was working on #1244, thanks for opening.

I like renaming the column to "Streaming". In English we can continue the phrasing nicely with options "For free" and "With purchase"

PR #1950, opened 2026-04-21
refactor : pre order / publish flow

### LINKED ISSUE #1894: Clarify the role of the "published at" and "released at" dates.
Image

This might need to be rethought generally UX-wise, but to users it has been unclear that the "publishedAt date" and the "releaseDate" map to go live and available. 

publishedAt -> means it's available for pre-order (and that `isPreview` songs are downloadable)
releaseDate -> means it's generally available for download. 

> Yea something more explicit. My expectation with workflow is to create / set-up album, then be confronted with three paths:
> 
> 1) have it set up on the backend but not publicly visible / accessible
> 
> 2) enable preorder
> 
> 3) enable sales / release the album

I'm personally not sure if this should happen at the end of the album form because it feels more top level metadata but I do think the UX aspect of that part of it could be more sequential.

LouisLouiseKay:
> Heads up: this PR is based on #1949  — it uses the useBulkSetTracksIsPreviewMutation hook introduced there. 

OK this is the big one, I tried to be super thourough but I'm sure I've missed some things so please check all of this.

## Summary

Adds a proper pre-order feature for albums. Before this, "pre-order" was just derived logically from `releaseDate > now()` in a few places on the frontend, which meant the state couldn't really be controlled by the artist — it would silently flip just because a date passed. I've turned it into an explicit `isPreorder` flag stored on the trackGroup, with a dedicated management UI, notifications, and an auto-end cron when the release date arrives.

The UI lives in a new `release-flow-2/:trackGroupId` route so we can beta-test it on production without touching the existing release flow. The backend changes are safe to merge on their own.

## Why an explicit flag instead of the date comparison

Two reasons:

1. Artists can make mistakes with dates, and these should be handled with the least amount of friction — a toggle is clearer than a date for that.
2. Pre-order should be an on-off switch artists can toggle themselves, without worrying about any other setting. The release date is now just used to schedule the auto-end, not to define whether something is a pre-order.

## What's in this PR

The 6 commits are ordered so the app stays working at every step (migrations first, backend next, then frontend on top of updated types):

1. **Prisma migrations** — adds `isPreorder`, `scheduleEndOnReleaseDate`, `makeTracksPreviewableOnRelease` on TrackGroup, makes `releaseDate` nullable, adds `NEW_ARTIST_PREORDER` to the NotificationType enum.
2. **Backend** — `endPreorderForTrackGroup` util (atomic Prisma transaction), `POST endPreorder` endpoint, `PUT trackGroups/{id}` accepting the new fields, `publish.ts` sending the right notification type, and the `end-scheduled-preorders` cron job.
3. **Frontend types + mocks** — the new fields and notification type added to `TrackGroup` and the test mocks.
4. **Pre-order management UI in release-flow-2** — PreOrderSection, SaveDraftBar, PublishButton updates, plus reworking all the frontend checks to use `isPreorder` instead of comparing the release date. Also fixes a bug where opening the edit page would silently disable `scheduleEndOnReleaseDate`.
5. **Notification feed** — followers get a `NEW_ARTIST_PREORDER` notification ("Available for Pre-order") when a pre-order is published, instead of the regular album notification.
6. **Cypress E2E tests** — 10 new tests covering the toggle, schedule, free-listen on release, end pre-order modal, button text, and download blocking.

## Testing

- [ ] Go to `/manage/artists/:id/release-flow-2/:trackGroupId` on an album
- [ ] Toggle "Setup as pre-order" on — all tracks switch to must-own, can be published without a release date
- [ ] Publish the album — followers get a "Available for Pre-order" notification in their feed
- [ ] Set a release date + enable "Schedule end on release date" + wait (or set date to 1 min in the future) → pre-order auto-ends via cron, tracks become listenable if "make tracks free-listen on release" was on
- [ ] Manually end a pre-order via the "Release pre-order now" modal → same effect
- [ ] As a user who pre-ordered: "Pre-ordered" disabled button shown, no download until release
- [ ] Toggle pre-order off → tracks reset to free-listen, schedule flags reset

## Not in there

- Cancel pre-order with Stripe refunds (we'll need to think that one through properly, have a few ideas for a temporary option sittin in another PR once we've validated that this works)
- The old release flow is unchanged — this adds a parallel beta route on purpose
- Disable autosave on published albums (working on this, same, we'll follow up once we've made sure this is working)

How it looks right now


client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/AlbumFormContent.tsx
```diff
@@ -13,25 +13,36 @@ import { css } from "@emotion/css";
...
 
   const { artistId, trackGroupId } = useParams();
   const urlSlug = watch("urlSlug");
+  const releaseDateValue = watch("releaseDate");
+
+  const hasReleaseDate = Boolean(releaseDateValue);
 
   return (
     <>
+      {isFlowV2 && <SaveDraftBar reload={reload} />}
```
simonv3: Oh, I thought you'd be creating a duplicate <AlbumFormContent> to avoid conflict with the existing functionality.

test/routers/artists/{id}.feed.spec.ts
```diff
@@ -464,7 +464,7 @@ describe("artists/{id}/feed", () => {
       );
       assert.equal(
         createItem.object.published,
-        trackGroup.releaseDate.toISOString()
+        trackGroup.releaseDate!.toISOString()
```
simonv3: I think there's specific lint rules to not use this pattern because it can have unexpected logic consequences?

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/AlbumFormContent.tsx
LouisLouiseKay: Ah! Should I?

LouisLouiseKay:
An important remark before we remove the old flow : 

Currently people who have set a publication date have basically "scheduled publication" of their album. This PR won't impact that behavior, however, since I've removed scheduling of publication for now, it means if we remove the old flow before the new one has a scheduler, these people won't be able to access their publication date anymore.

The way I see the scheduler is basically that we add either an icon (to be defined) or litterally a "schedule publication" button near the big publish button, and the scheduled publication date is set in a modal, and once it's set, the form indicates "scheduled to be published at : date". The manual publish button stays available for manual publication, but we might add a warning "are you sure you want to publish now? This release is scheduled to be released on : date".

timglorioso:
> The way I see the scheduler is basically that we add either an icon (to be defined) or litterally a "schedule publication" button near the big publish button

Yeah I was imagining a separate button that displays a menu with:
- Schedule...
- Unpublish

like this kind, except not checkboxes, just a normal menu:

Screenshot 2026-04-21 at 1 55 26 PM

LouisLouiseKay:
I don't know if we need a menu for unpublish, for me it's an on/off state the button either indicates "publish" or "unpublish" (currently "make private" which I think is clearer but maybe not)? I did think of putting the scheduler button kind of in the same spot as in the picture, but with a clock icon or something. And then the pop-up modal would be to make sure the feature is clear and it's "out of the context of the rest of the form" (it's basically just "schedule hitting that publish button" instead of "here's another weird date field" which is actually what the code does)

timglorioso:
> I don't know if we need a menu for unpublish,

Just want to clarify that my suggestion is for unpublish/make private to be a menu item, not a separate menu

something like this
Screenshot 2026-04-21 at 2 19 01 PM

client/src/translation/en.json
```diff
@@ -949,8 +948,28 @@
     "goal": "Goal",
     "now": "now",
     "publish": "Publish release {{ date }}",
+    "publishSimple": "Publish",
+    "makePrivateSimple": "Make private",
     "publishedSuccess": "Successfully published release, it's now publicly available.",
     "publishPreorder": "Publish pre-order now",
+    "preOrder": "Pre-order",
+    "setupAsPreorder": "Setup as a pre-order",
+    "preorderTrackInfo": "When you setup your release as a pre-order, <strong>your tracks are automatically made non-streamable by default during the pre-order campaign.</strong> Head over to the track section and replace \"must own\" with \"free listen\" on the tracks you would like to be previewable during that time, these tracks will also be downloadable by those who pre-order your release.",
```
simonv3, line 967:
```suggestion
    "preorderTrackInfo": "When you setup your release as a pre-order, <strong>your tracks are automatically made non-streamable by default during the pre-order campaign.</strong> Head over to the track section and replace \"not playable\" with \"playable\" on the tracks you would like to be previewable during that time, these tracks will also be downloadable by those who pre-order your release.",
```

client/src/translation/en.json
```diff
@@ -949,8 +948,28 @@
...
+    "schedulePreorderReleaseDate": "Schedule the pre-order campaign to end on release date",
+    "releaseDatePreorderHint": "When the release date arrives, the pre-order campaign will automatically end and the release becomes available for download to those who pre-ordered. Be careful that the date is correct and everything is in order on the release page! When the pre-order campaign ends, a notification will be sent to your followers and subscribers that the album is now available, and people who pre-ordered will receive a download link to the full release.",
+    "noReleaseDateSet": "You need to set a release date first in the Key Details section before you can schedule the end of the pre-order campaign.",
+    "schedulePreorderNotPublicationHint": "<strong>Attention!</strong> This doesn't schedule your album's publication — it only controls when it transitions from \"pre-order\" to \"released\". You'll still need to publish the album manually to make it visible to your listeners.",
```
simonv3:
```suggestion
    "schedulePreorderNotPublicationHint": "<strong>Attention!</strong> This doesn't schedule your album's publication! It only controls when it transitions from \"pre-order\" to \"released\". You'll still need to publish the album manually to make it visible to your listeners.",
```

client/src/translation/en.json
```diff
@@ -949,8 +948,28 @@
...
+    "makeTracksPreviewableOnRelease": "Set all tracks as playable on release day",
+    "setTracksAsFreeListen": "Set all tracks as playable on release day",
+    "endPreorderCampaign": "Release pre-order now",
+    "releasePreorderHint": "Click this button when you're ready to end the pre-order campaign if you prefer to activate that manually on release day (Have a particular time you want to release at? You'd rather be around for the release moment? There are many valid reasons you might prefer it that way).",
```
simonv3:
```suggestion
    "releasePreorderHint": "Click this button when you're ready to end the pre-order campaign if you prefer to activate that manually on release day (Have a particular time you want to release at? You'd rather be around for the release moment?).",
```

client/src/components/ManageArtist/PublishButton.tsx
```diff
@@ -139,16 +134,36 @@ const PublishButton: React.FC<{
         isLoading={isPublishing}
         onClick={publishTrackGroup}
         disabled={isPublishing}
+        className={css`
```
simonv3: Can this use tailwind?

client/src/components/ManageArtist/PublishButton.tsx
```diff
@@ -111,17 +112,11 @@ const PublishButton: React.FC<{
...
 
-  const beforeReleaseDate = new Date(trackGroup.releaseDate) > new Date();
-
-  const publishButton = beforeReleaseDate ? "publishPreorder" : "publish";
-  const privateButton = beforeReleaseDate
-    ? "makePreorderPrivate"
-    : "makePrivate";
-
   return (
     <div
```
simonv3: Convert to tailwind please

client/src/components/ManageArtist/PublishButton.tsx
```diff
@@ -139,16 +134,36 @@ const PublishButton: React.FC<{
...
+        className={css`
+          background-color: var(--mi-green-500) !important;
+          border-color: var(--mi-green-700) !important;
+          color: var(--mi-green-100) !important;
+
+          svg {
+            fill: var(--mi-green-100) !important;
+          }
+
+          &:hover:not(:disabled) {
```
simonv3: It's odd to me that this button is clickable but so disabled looking. 

Image

Is this an accessibility thing? I'm also wondering that we need to wait to determine if there's a cover from the API? 

What's happening is I haven't uploaded a cover yet. I click on the Publish button. It spins a while. Then I get a snackbar error saying that it needs a cover. I think the UI could have determined this earlier?

simonv3: Oh actually I think that's just the color of the publish button? It looks that way when everything is fine too. And the Make Private looks similarly muted. This is a new artist and I haven't set any colors on it yet.

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PreOrderSection.tsx
```diff
@@ -0,0 +1,268 @@
...
+    <FormSection>
+      <h2>{t("preOrder")}</h2>
+      <div className="flex flex-col gap-4">
+        <Toggle
+          label={t("setupAsPreorder")}
+          toggled={isPreorder}
+          onClick={handlePreorderToggle}
+        />
+
+        {isPreorder && (
```
simonv3, line 172: Because this is driven by the API it takes a moment to appear. Could we add a loading indicator for more immediate feedback or assume it went through and set the isPreorder flag sooner?

Image

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PreOrderSection.tsx
```diff
@@ -0,0 +1,268 @@
...
+        artistId: Number(artistId),
+      });
+      reload();
+    } catch (e) {
+      errorHandler(e);
+      reload();
+    }
+  };
+
+  const handlePreorderToggle = async (enabled: boolean) => {
```
simonv3, line 101: This is super niche but there's a slightly odd race condition happening in the form. 

If a user edits the release date, and then clicks "set up preorder" before the release date is finished saving, it returns this pretty cryptic message and also seems to reset the release date? 

Image

simonv3 (changes requested):

client/src/translation/en.json
simonv3: Kind of a suggestion but I thought this would be included in this PR, unless it's a different work thing, in which case, ignore!

LouisLouiseKay: I was thinking of puting that in another PR separately once we'd sure all of this worked properly ! I could include it I guess, the only change I've made to playable / non playable is in the "set all tracks as playable" (because it's a "hands off" feature you don't have go look for the specific toggle so it's ok if it doesn't match the exact wording in track section).

client/src/components/ManageArtist/PublishButton.tsx
LouisLouiseKay: This is because you probably (weirdly actually?) don't have PR1 on the branch you're on that adds the needed color variable in the style css

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PreOrderSection.tsx
LouisLouiseKay: Actually I had one and removed it because it was very quick on my end and it felt a bit weird but absolutely no problem to add it back yeah, it makes sense on slower connexions.

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PreOrderSection.tsx
LouisLouiseKay: Oh yeah, I think I have an idea of what's going on I'll look into it, we want to bulletproof those kind of odd cases actually.

client/src/components/ManageArtist/PublishButton.tsx
LouisLouiseKay: It's in #1948

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PreOrderSection.tsx
LouisLouiseKay: So actually (and this is more of a "album form wide" issue) because anytime something reloads it can override all fields with the previously stored value, we have *several* spots in the code where this can potentially happen. The fix is to save the temporary value. I think this could just be resolved by changing how `methods.reset` overwrites any field the user has just touched.

```suggestion
    methods.reset(defaultValues, { keepDirtyValues: true });
```

I think this would actually fix a bunch of other odd behaviors of the same type because that race wasn't just happening across date, say you change your title and then click "schedule", it similarly reloads the entire page and reset saved values.

client/src/components/ManageArtist/PublishButton.tsx
simonv3: The issue is that this branch hasn't been rebased on that one? I just checked out the branch as it is existing on here.

client/src/translation/en.json
timglorioso: We were also considering the field to be "Allow streaming" and the two current options to be "For free" and "With purchase". See https://github.com/funmusicplace/mirlo/issues/1245

client/src/translation/en.json
timglorioso: My opinion is that exclamation marks should be used very sparingly in ux copy, personally I'd keep the dash or replace with a period.

simonv3: Yeah a period would be fine. I'm just wary of emdash personally because the wider world now associates them (regretably, imo, for the world of writing, cause i love me an emdash) with AI haha.

client/src/components/ManageArtist/PublishButton.tsx
LouisLouiseKay: Yeah exactly

client/src/translation/en.json
LouisLouiseKay: For what it's worth I don't have strong opinion (yes ! IT HAPPENS !) Emdash, dots, exclamation point, all good for me. So let's go with a period? I think it's otherwise clearly signified enough that people must double triple quadruple check those settings before release in the wording.

timglorioso: I guess I shouldn't be surprised others have noticed—I didn't know that was a thing—now I know

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PreOrderSection.tsx
```diff
@@ -0,0 +1,268 @@
...
+                    id="hint-schedule-preorder-release"
+                    className="block mt-1"
+                  >
+                    <Trans
+                      i18nKey="manageAlbum.schedulePreorderNotPublicationHint"
+                      components={{ strong: <strong /> }}
+                    />
+                  </small>
+                </div>
+              </div>
```
timglorioso, line 204: Chef's kiss!

LouisLouiseKay: I hoped you'd like that one !

client/src/components/ManageArtist/PublishButton.tsx
LouisLouiseKay: OK rebased on main should be all good

client/src/translation/en.json
LouisLouiseKay: That's the reason I didn't put this in this PR ahah, I know it's like, a can of worm to think of the terminology here. I like "playable / non playable" because it's tech agnostic and plain language but yeah there are many ways to skin this cat and I think we can get back to it in a simple PR where we just find the correct wording after we merged all this headache inducing mess.

LouisLouiseKay: Marking this as revolved for now, but you can add "..." and "to be continued" with a bit of tense music.

client/src/components/TrackGroup/PurchaseOrDownloadAlbumModal.tsx
```diff
@@ -106,9 +111,26 @@ const PurchaseOrDownloadAlbum: React.FC<{
...
         )}
-        {/* {showPurchase && (
-          <AddToCollection trackGroup={trackGroup} fixed={fixed} />
-        )} */}
+        {showPreOrdered && (
+          <ArtistButton
+            variant="outlined"
+            disabled
+            aria-disabled="true"
+            className={css`
```
simonv3, line 119: For the future, tailwind please

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/SaveDraftBar.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const [isSaving, setIsSaving] = React.useState(false);
+
+  const isPublished = isTrackGroupPublished(existingObject);
+
+  const handleSaveDraft = async () => {
+    if (!trackGroupId || !artistId) return;
+    setIsSaving(true);
+    try {
+      const values = getValues();
+      await api.put(`manage/trackGroups/${trackGroupId}`, {
```
simonv3, line 33: Not for this PR but we should wrap this in a query and use that to manage caching and cache clearing rather than passing the `reload` function around.

simonv3 (dismissed):

simonv3 (approved):

PR #1954, opened 2026-04-22
Improve accessibility of merch, subscription tier, post, profile forms

timglorioso:
See commits for summary of changes.


client/src/components/ManageArtist/Posts/EditPostHeader.tsx
```diff
@@ -28,58 +28,61 @@ const EditPostHeader: React.FC<{
...
+              artistId={artist.id}
+              reload={reload}
+              onClose={onClose}
+            />
+          )}
+          <PublishPostButton post={post} reload={() => refetch()} />
+        </div>
       </div>
-    </div>
+      <h1>{post.title || t("untitledPost")}</h1>
```
timglorioso, line 84: Generally need to revise commits to ensure translation keys are present

client/src/components/Artist/Artist.tsx
```diff
@@ -142,64 +144,86 @@ function Artist() {
...
+              </NavLink>
+            </li>
+            <li className="tab-secondary">
+              <ArtistButtonQuickLink
+                ariaLabel={t("editRoster")}
+                to="profile/label"
+                icon={<FaEdit />}
+              />
+            </li>
+          </>
```
timglorioso: This is not the ideal solution, but it's better than before. Icon links have labels and are distinct list items.

client/src/components/Artist/Artist.tsx
```diff
@@ -142,64 +144,86 @@ function Artist() {
     <>
```
timglorioso, line 150: TBH this should probably be `nav` with `aria-label` set to something like "artist" so the announcement is like "artist navigation".

client/src/components/Artist/Artist.tsx
```diff
@@ -142,64 +144,86 @@ function Artist() {
...
-          </li>
+          <>
+            <li className="tab-primary">
+              <NavLink to="roster">
+                {artist.properties?.titles?.roster || t("roster")}
+              </NavLink>
+            </li>
+            <li className="tab-secondary">
+              <ArtistButtonQuickLink
+                ariaLabel={t("editRoster")}
```
timglorioso: These translations need to be interpolated with any custom names.

simonv3 (dismissed):
Looks great, thanks Tim!

PR #1973, opened 2026-04-28
Artist colors expansion

LouisLouiseKay:
Follow-up to my previous artist color migration PR (#1970) which renamed the database keys to button / buttonText / text / background. This one brings the CSS variable layer in sync with the same naming, and adds the things that were waiting on the rename:

- A new optional secondaryText color slot for caption / meta text. Falls back to text when not set.
- An adaptive tint and contrast color system, calibrated by the artist's background brightness. Light bg gets a black tint and a pure black contrast color, dark bg gets the opposite. Fixes the cases where darken / lighten didn't work on artists with very light or very dark themes.
- A button-calibrated variant of the same, so surfaces that paint a tint behind button-color text (header strip, post cards, subscription tier cards) always read well (we can still easily adjust intensity in the future).
- A new transparentContainer toggle on the artist properties. Replaces the pre-existing "undefined background = transparent" hack so artists keep their bg image showing through. A migration backfills it to true for every artist who relied on the hack, no manual intervention needed.
- ArtistColorsProvider also gets rewired along the way so the form preview can update instantly without a save / reload roundtrip.

I'll probably add a more refined bunch of improvements for how the layout must behave when the transparent toggle is on but I'll do that in a separate PR this one was a lot already.


client/src/components/common/Button.tsx
```diff
@@ -25,9 +26,18 @@ const CustomButton = styled.button<Sizable>(
...
-    let cssColorVariable = `--mi-${buttonRole ?? "primary"}-color`;
-    var primaryColor = bodyStyles.getPropertyValue(cssColorVariable);
-    let secondaryColor = bodyStyles.getPropertyValue(`--mi-secondary-color`);
+    // buttonRole "primary" maps to the artist accent (--mi-button-color in the
+    // new scheme; --mi-button-color in legacy), other roles keep their
+    // explicit Mirlo system var. Contrast color for filled buttons follows
+    // the same convention as ArtistButton: --mi-button-text-color.
+    const cssColorVariable =
+      buttonRole && buttonRole !== "primary"
+        ? `--mi-${buttonRole}-color`
```
simonv3, line 35: Won't this clash with people's designs? Just thinking also that here we're conflating button role with button color. Makes it a little harder to reason about. Don't know that that's part of this PR though.

simonv3 (approved):

PR #1977, opened 2026-04-29
Add private public visibility state

LouisLouiseKay:
Sort of a follow up to the "pre-order flow" rework, this adds a public/private flag on trackgroups so artists can publish a release that's only accessible by sharing the URL, without it being listed publicly anywhere on Mirlo, search, RSS feed or follower notification. Defaults to public so existing albums stay visible (if they were "made private" prior to that, it'll be a transparent change to the artists cause they'll just see the Publish button like they used to).

The release management gets a "Publish as" set of radio buttons (with info text that adapts to the status of the release) and an "Update release" button once the album is Published. The artist's own private releases stay visible to them on their public page, grayscaled with a lock and "Only visible to you" caption to avoid confusion.

Follower notifications now only get sent once per release on first publish or first private to public flip, with a fallback to make sure that existing albums never re-notify either.

Normally I've made sure the legacy v1 manage flow is untouched.


client/src/components/Artist/ArtistTrackGroup.tsx
```diff
@@ -88,34 +91,54 @@ const ArtistTrackGroup: React.FC<ArtistTrackGroupProps> = ({
   ...props
 }) => {
   const length = size === "small" ? 300 : 600;
+  const { user } = useAuthContext();
+  const { t } = useTranslation("translation", { keyPrefix: "manageArtist" });
+  const isPublished =
+    !!trackGroup.publishedAt && new Date(trackGroup.publishedAt) <= new Date();
```
simonv3: isTrackGroupPublished?

src/routers/v1/artists/{id}/index.ts
```diff
@@ -24,12 +25,21 @@ export default function () {
       const parsedId = await findArtistIdForURLSlug(id);
       let isUserSubscriber = false;
       if (parsedId) {
+        const ownership = await prisma.artist.findFirst({
```
simonv3: Should labels be able to see it?

src/routers/v1/manage/trackGroups/{trackGroupId}/index.ts
```diff
@@ -81,9 +84,12 @@ export default function () {
...
 
       return res.status(200).json({
-        result: processor.single(trackGroup, {
-          loggedInUserId: loggedInUser.id,
-        }),
+        result: {
+          ...processor.single(trackGroup, {
+            loggedInUserId: loggedInUser.id,
+          }),
+          hasNotifiedFollowers: trackGroup.notifiedFollowersAt !== null,
```
simonv3: Could this be included in the processor.single? That's the point of it!

client/src/components/Artist/ArtistTrackGroup.tsx
LouisLouiseKay: Oh yeah

src/routers/v1/artists/{id}/index.ts
LouisLouiseKay: Oh yeah they definitely should.

src/routers/v1/manage/trackGroups/{trackGroupId}/index.ts
LouisLouiseKay: Done

src/routers/v1/artists/{id}/index.ts
LouisLouiseKay: OK I think this is fixed in a new commit

simonv3 (dismissed):

simonv3 (approved):

PR #1987, opened 2026-04-30
fix: hide Roster tab on label profile when roster is empty

### LINKED ISSUE #1629: Don't display Roster for artists if there are no artists on the label's roster
If the array of artists is blank, we shouldn't be showing the tab in the header.

Image

peterklingelhofer:
Closes #1629

### DIFF
--- a/client/src/components/Artist/Artist.tsx
+++ b/client/src/components/Artist/Artist.tsx
@@ -157,7 +157,7 @@ function Artist() {
       >
         <ArtistTabs>
           {artist.isLabelProfile &&
-            (artist.user?._count?.artistLabels ?? 0) > 0 && (
+            (artist.user?.artistLabels?.length ?? 0) > 0 && (
               <>
                 <li className="tab-primary">
                   <NavLink to="roster">{rosterTitle}</NavLink>
--- a/client/src/components/Artist/Artist.tsx
+++ b/client/src/components/Artist/Artist.tsx
@@ -156,20 +156,21 @@ function Artist() {
         }
       >
         <ArtistTabs>
-          {artist.isLabelProfile && (
-            <>
-              <li className="tab-primary">
-                <NavLink to="roster">{rosterTitle}</NavLink>
-              </li>
-              <li className="tab-secondary">
-                <ArtistButtonQuickLink
-                  ariaLabel={t("editTitled", { title: rosterTitle })}
-                  to="/profile/label"
-                  icon={<FaEdit />}
-                />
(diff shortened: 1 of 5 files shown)


simonv3 (dismissed):
Thanks! Looks like @LouisLouiseKay merge caused conflicts

src/utils/artist.ts
```diff
@@ -591,6 +591,9 @@ export const singleInclude = (queryOptions?: {
     user: {
       select: {
         currency: true,
+        _count: {
+          select: { artistLabels: true },
```
simonv3: should it select only non-soft deleted deleted artistLabels?

client/src/components/Artist/Artist.tsx
```diff
@@ -156,20 +156,21 @@ function Artist() {
...
-                <ArtistButtonQuickLink
-                  ariaLabel={t("editTitled", { title: rosterTitle })}
-                  to="/profile/label"
-                  icon={<FaEdit />}
-                />
-              </li>
-            </>
-          )}
+          {artist.isLabelProfile &&
+            (artist.user?._count?.artistLabels ?? 0) > 0 && (
```
simonv3: is there a way to make this a little easier on the eyes api-wise? Do we need a count or just a length check?

simonv3 (commented):
two questions

src/utils/artist.ts
peterklingelhofer: Addressed here: https://github.com/funmusicplace/mirlo/pull/1987/changes/fadb224a717ce276c170d3abf0a786cd1c5811cc

client/src/components/Artist/Artist.tsx
peterklingelhofer: Addressed here: https://github.com/funmusicplace/mirlo/pull/1987/changes/fadb224a717ce276c170d3abf0a786cd1c5811cc

simonv3 (approved):

xyzzyestudioweb:
Sorry, there’s still a small detail that hasn’t been resolved yet. Although it’s true that the roster tab no longer appears when there are no artists in it, it still happens that when I access the main label profile page, the URL redirects me to “roster” and shows an empty space. I’m attaching a screenshot.

Maybe this could be solved with an option where you can specify which tab you want to use as the main profile tab, and then generate the redirect based on that.

[https://mirlo.space/ediciones-subterranias/](https://mirlo.space/ediciones-subterranias/)

image

PR #1991, opened 2026-04-30
fix(client): default merch home country to the artist location

### LINKED ISSUE #1974: Default merch destination to the one of the label.
> Also, small suggestion. When I go to create merch, it auto selects the first destination (Afghanistan) as the home destination for shipping. It'd be cool if the destination that appeared was the location set for the label (or artist if using their own page for merch).

peterklingelhofer:
Closes #1974

### DIFF
--- a/client/src/components/ManageArtist/Merch/MerchDestinations.tsx
+++ b/client/src/components/ManageArtist/Merch/MerchDestinations.tsx
@@ -31,7 +31,20 @@ const currencyToCountryMap = countryCodesCurrencies.reduce(
   {} as { [key: string]: any }
 );
 
-const MerchDestinations: React.FC<{}> = () => {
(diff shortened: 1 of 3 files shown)


client/src/components/ManageArtist/Merch/MerchDestinations.tsx
```diff
@@ -42,13 +55,26 @@ const MerchDestinations: React.FC<{}> = () => {
     queryManagedMerch(merchParamId ?? "")
   );
 
+  const artistCountryCode = React.useMemo(() => {
```
simonv3, line 58: Hmm, this is interesting. I guess it's fine. I wonder if people will notice / complain / have weird edge cases.

simonv3 (approved):

PR #1996, opened 2026-05-01
feat: hydrate public Post page via parseIndex injection

### LINKED ISSUE #1981: Hydrate posts into page from parseIndex
Using the same structure as the user hydration in parseIndex, can we hydrate posts so that they display right away without flashing a load for a moment?

* [ ] Move the post.single function into the serialize folder to have everything in one place
* [ ] Use the serialize function in parseIndex to inject meta data in the same format as the user on the page.
* [ ] Preload that metadata into the public Post component in the same way as user profile gets loaded in.

peterklingelhofer:
Closes #1981

### DIFF
--- a/src/parseIndex.ts
+++ b/src/parseIndex.ts
@@ -41,6 +41,7 @@ type RouteContext<T extends RouteParams = RouteParams> = {
   client: Client;
   avatarUrl?: string;
   params: T;
+  req?: Request;
 };
 
 type RouteHandler<T extends RouteParams = RouteParams> = (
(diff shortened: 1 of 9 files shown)


src/parseIndex.ts
```diff
@@ -781,6 +834,15 @@ export const analyzePathAndGenerateHTML = async (
     const routeParams = matchRoutePattern(segments);
     if (routeParams) {
       await dispatchRoute(routeParams, { $, client, avatarUrl });
+
+      // For a specific post, hydrate the page with the serialized post so the
+      // public Post component renders immediately without a load flash.
+      if (
```
simonv3: Can't this be put in the `handlePost` function on line 305?

simonv3 (approved):

PR #1997, opened 2026-05-01
feat: store client id on track play

simonv3:


src/auth/cors.ts
```diff
@@ -83,6 +83,9 @@ export const corsCheck = async (...args: [Request, Response, NextFunction]) => {
         log.info(
           `Found ${clients.length} clients with that API key: ${clients.map((c) => `${c.applicationName}: ${c.allowedCorsOrigins.join(", ")}`).join(", ")}`
         );
+        if (clients.length === 1) {
```
peterklingelhofer, line 86: `clients.length > 1` should be impossible right? Could add a throw case for that beneath this if clause I suppose just in case something wild occurs?

peterklingelhofer (approved):
Makes sense to me, just one comment/nit, approving regardless

src/auth/cors.ts
simonv3: tbh the logic of this function is a little weird and I'm not sure it makes sense

PR #2005, opened 2026-05-02
Refactor embed player

LouisLouiseKay:
This was initially just going to be a "fix" but it's better reframed as a refactor because there was just... too much stuff to fix. Here's the list of changes :

- Trackgroup and label widgets visually reworked to match the single-track widget style. On mobile, trackgroup widget shows the cover + title up top with the tracklist below. On desktop, tracklist is on the side.
- Audio progress bar now supports drag and drop scrubbing, plus keyboard navigation (arrow keys to step, Home/End to jump to start/end).
- Buy and Support/Follow buttons added to widgets that take the viewer to the artist's release page (with the purchase modal opening if you clicked buy) or to their support page.
- Track widget now shows the current playback time next to the play button.
- Trackgroup widget shows the track count next to the artist name.
- Fade overlay at the bottom of the tracklist hints at more tracks when it overflows (uses the same system than the Subscription Tiers carrousel except top down instead of right left).
- Embedded widgets stay in sync with the main Mirlo player on play/pause/current time when embed on Mirlo (the hardest part of the work by far, that thing was always broken and it was a headache).
- New "pill" button variant added to the common Button component for small outlined buttons (because we didn't have anything more compact than the base compact button and those were still way too big, plus I think pill buttons might be useful elsewhere in the UI now with all the features we've added).
- Various smaller fixes: target="_blank" attribute typo that I'd left on widget links, missing rel="noopener noreferrer", removed a bunch of "!important" css here and there etc.


client/src/components/Player/AudioWrapper.tsx
```diff
@@ -226,39 +229,42 @@ export const AudioWrapper: React.FC<{
...
-          if (data.details === Hls.ErrorDetails.MANIFEST_LOAD_ERROR) {
-            if (data.networkDetails?.responseText) {
-              if (
-                data.networkDetails.responseText.includes(
-                  "Track play limit exceeded"
-                ) &&
-                !hasShownBuyModalBeenShown
-              ) {
-                setHasOverplayedSong(true);
+      <React.Suspense fallback={null}>
```
simonv3, line 232: Why?

client/src/utils/playerSync.ts
```diff
@@ -0,0 +1,131 @@
+import React from "react";
```
simonv3, line 1: I feel like we should extract this into a library tbh. Feels like it would be useful either way.

simonv3 (approved):

Issue #1403, opened 2025-08-07
People still able to upload mp3s for some reason


simonv3:
In the database, the audio extension is still set to .mp3 for some uploaded tracks, which shouldn't be possible.

PR #2007, opened 2026-05-02
feat(upload): show accepted file formats next to upload inputs

### LINKED ISSUE #1638: Provide a list of formats accepted for uploading
The formats of files that can be uploaded to Mirlo are not clearly specified. This is true for music files of course, but also for pictures and maybe other files elsewhere (covers, artist page background etc...). Once you click on the uploader and are browsing the files on your device you can indeed also see and pick accepted formats but to some this might be too late to wonder.

I suggest this list of accepted formats be introduced following this example: `Accepted file formats : [CoverFileFormats]`
This will allow each instance maintainer to change easily change the file formats accepted and translators to only care about the introductory sentence and not the file formats themselves (who do not need translation anyway).

Based on feedback on Discord:
https://discord.com/channels/1070731899317796974/1441261647221231687/1441395739283554387
> Hi! What file formats can I upload to Mirlo? I started with flac but it seemed to hang forever processing, so I deleted them all and tried wav instead. I am having difficulty finding any information about what file formats you accept for uploads. [...] Could be I am just impatient? But without any indication of what file formats are accepted, it is hard to know what to do.

peterklingelhofer:
Closes #1638

### DIFF
--- a/client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ReplaceTrackAudioInput.tsx
+++ b/client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ReplaceTrackAudioInput.tsx
@@ -8,6 +8,10 @@ import useJobStatusCheck from "utils/useJobStatusCheck";
 import api from "services/api";
 import { useSnackbar } from "state/SnackbarContext";
 import LoadingSpinner from "components/common/LoadingSpinner";
+import { formatAcceptList } from "utils/uploadFormats";
+
+const ACCEPTED_AUDIO =
+  "audio/mpeg,audio/flac,audio/wav,audio/x-flac,audio/aac,audio/aiff,audio/x-m4a";
 export interface FormData {
   trackFile: FileList;
 }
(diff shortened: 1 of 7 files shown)


client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/ReplaceTrackAudioInput.tsx
```diff
@@ -8,6 +8,10 @@ import useJobStatusCheck from "utils/useJobStatusCheck";
 import api from "services/api";
 import { useSnackbar } from "state/SnackbarContext";
 import LoadingSpinner from "components/common/LoadingSpinner";
+import { formatAcceptList } from "utils/uploadFormats";
+
+const ACCEPTED_AUDIO =
+  "audio/mpeg,audio/flac,audio/wav,audio/x-flac,audio/aac,audio/aiff,audio/x-m4a";
```
simonv3: Realizing that if we remove `audio/mpeg` from this list we'll probably also close https://github.com/funmusicplace/mirlo/issues/1403
```suggestion
  "audio/flac,audio/wav,audio/x-flac,audio/aac,audio/aiff,audio/x-m4a";
```

simonv3 (dismissed):
One recommendation if you want to include it to prevent an other bug.

simonv3 (approved):

PR #2010, opened 2026-05-02
fix: newsletter restyling

LouisLouiseKay:
First and foremost, the biggest point of this PR was to fix text being displayed in full bold weight in emails.

This was due to the "Avenir" font, which defaulted to "Avenir Black" depending on which system you were on (that's why emails worked on Mobile, and in other OS, but not on Windows). When switching to other fonts it all worked fine.

Also recentered images and text, and added some light styling to better differenciate things like blogpost title, "artist just published a new post", Unsubscribe and "sent using Mirlo".

### DIFF
--- a/emails/announce-post-published/html.pug
+++ b/emails/announce-post-published/html.pug
@@ -2,18 +2,22 @@ extends ../layout.pug
 
 block content
 
-    if post.featuredImage
-      div.mw600 
-        a(href!=`${client}/post/${post.id}`) 
-          img(src=post.featuredImage.src, alt="Featured Image", style="width:100%;height:auto;max-height:400px;object-fit:cover;margin-bottom:16px;")
-    p  
-    p #{artist.name} just published a new post on Mirlo: 
-    h2 #{post.title}
-    p 
-      a(href!=`${client}/${artist.urlSlug}/posts/${post.urlSlug}`) Read it on Mirlo.
-    div.mw600 
+    div.mw600.email-header
+      if post.featuredImage
+        a(href!=`${client}/post/${post.id}`)
+          img(src=post.featuredImage.src, alt="Featured Image", style="width:100%;height:auto;max-height:400px;object-fit:cover;margin-bottom:8px;")
+
+      p.email-byline #{artist.name} just published a new post on Mirlo
+      h1.email-title #{post.title}
+      p.read-on-mirlo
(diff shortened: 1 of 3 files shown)


emails/announce-post-published/html.pug
```diff
@@ -2,18 +2,22 @@ extends ../layout.pug
...
-          img(src=post.featuredImage.src, alt="Featured Image", style="width:100%;height:auto;max-height:400px;object-fit:cover;margin-bottom:16px;")
-    p  
-    p #{artist.name} just published a new post on Mirlo: 
-    h2 #{post.title}
-    p 
-      a(href!=`${client}/${artist.urlSlug}/posts/${post.urlSlug}`) Read it on Mirlo.
-    div.mw600 
+    div.mw600.email-header
+      if post.featuredImage
+        a(href!=`${client}/post/${post.id}`)
```
simonv3: This should probably link to the artist context of the post right? 


```suggestion
        a(href!=`${client}/{post.artist.urlSlug}/posts/${post.id}`)
```

emails/announce-post-published/html.pug
```diff
@@ -2,18 +2,22 @@ extends ../layout.pug
...
-      div.mw600 
-        a(href!=`${client}/post/${post.id}`) 
-          img(src=post.featuredImage.src, alt="Featured Image", style="width:100%;height:auto;max-height:400px;object-fit:cover;margin-bottom:16px;")
-    p  
-    p #{artist.name} just published a new post on Mirlo: 
-    h2 #{post.title}
-    p 
-      a(href!=`${client}/${artist.urlSlug}/posts/${post.urlSlug}`) Read it on Mirlo.
-    div.mw600 
+    div.mw600.email-header
```
simonv3: Do we have to set max width on every element? Putting it on a parent element doesn't work?

LouisLouiseKay: I think we could, to be honest I tried to keep a maximum of the layout as it already was and not change it too much and that's what was already set. Ideally I think I'll actually try to nest two elements so we can add background color.

emails/announce-post-published/html.pug
LouisLouiseKay: Ah, I didn't touch the link references, thought they worked already ! I copy pasted the url of "read it on mirlo" which I remembered working.

simonv3 (approved):

Issue #1278, opened 2025-06-25
Missing translation in the "About Us" page


Cwpute:
Text is translated (and translatable in Transifex) up to "Price & Features", after that it's all untranslated (and untranslatable on Transifex).
I usually find things like that, would i be able to do it myself with my limited coding knowledge ?

Cwpute:
What would be managed by the third-party exactly ?

simonv3:
So I think it would be an external link, like a gitbook or something, for documentation. Right now pushing a code update to update our FAQ or user how-to is not ideal. It would make maintaining that more accessible to users rather than to just us developers.

PR #2023, opened 2026-05-03
fix: convert part of About page to i18n (remaining hardcoded text)

git-ksupriya:
Some parts of the About page still contain hardcoded text 
instead of using i18n (`t` / `Trans`).

This PR converts the "Our story" section as a first step.

A previous issue (#1278) addressing this appears to be closed,
but some sections are still untranslated.

Happy to extend this if needed :)


simonv3:
Hi @git-ksupriya Thanks for the PR. The reason that issue is closed is because we've moved our about page to the docs section. The mirlo.space/pages/about redirects there now so this component isn't actually used and tbh should be removed.

git-ksupriya:
Got it, makes sense.. I wasn’t aware it had been moved to docs. Closing this PR since the component is no longer in use

simonv3:
@git-ksupriya thanks though! Curious how you found this to work on. Were you browsing the code?

git-ksupriya:
Yeah, I was browsing through the repo and issues to understand the codebase, looking for a small beginner-friendly thing to work on.. and noticed the About page still had some hardcoded sections. Saw the older issue related to it and tried converting one section to i18n as a small contribution

PR #2031, opened 2026-05-04
Don't display deleted artists on label

### LINKED ISSUE #1641: Label page/admin displays deleted artists
### Observed behavior

Label page/admin displays deleted artists.

### Expected behavior

Label page/admin does not display deleted artists.

### Steps to reproduce

1. Log in
2. Enable label on account
3. Create or use existing artist
4. Add artist to label roster
5. Confirm by artist
6. Delete artist

timglorioso:
Close #1641


timglorioso (commented):
Leaving review

timglorioso:
GitHub is being buggy but I wanted to comment on this: https://github.com/funmusicplace/mirlo/pull/2031/changes#diff-38ee03d3dddd37d5f94726b1314b28c9277a9827a9425e4e3639bcbd4e80d6fbR74

I'm not sure what to do here. If user doesn't exist, what does that mean for the fetched artist? Is that even possible? How to know if possible? I thought about returning a 500, right now it silently does nothing.

src/routers/v1/artists/{id}/index.ts
```diff
@@ -54,13 +54,48 @@ export default function () {
...
         isUserSubscriber = await checkIsUserSubscriber(loggedInUser, parsedId);
 
-        return res.json({
-          result: processSingleArtist(
-            artist as any,
-            loggedInUser?.id,
-            isUserSubscriber
-          ),
-        });
+        if (artist.isLabelProfile) {
```
timglorioso: @simonv3 I added this back and made it return something. Does this seem okay in relation to recent AP work?

client/src/components/Label/Roster.tsx
```diff
@@ -2,38 +2,40 @@ import { useTranslation } from "react-i18next";
...
     return <div>{t("labelNotFound")}</div>;
   }
 
-  const { data: label, isPending } = useQuery(queryLabelBySlug(artistId));
+  const { data: artistAsLabel, isFetching } = useQuery(
+    queryArtist({ artistSlug: slug })
+  );
 
-  if (isPending) {
+  if (isFetching) {
```
simonv3: If I understand my useQuery loading states correctly this will display the loading blocks every time we're fetching data, which is not just on page load but also every time useQuery decides that the cache can be updated because it's stale, which includes when you navigate back tot he page, or every couple of minutes. I think it doesn't look that good. Any reason specifically for doing this on isFetching rather than isPending?

src/routers/v1/artists/{id}/index.ts
simonv3: This just inserts the artistLabels object into the artist object from the user.artistLabels? The original singleInclude doesn't fetch this? Might be useful to add a comment stating the reasoning for this?

client/src/components/Label/Roster.tsx
timglorioso: I think it's worse to show something stale (e.g. deleted artist) and then have it disappear a second later.

The query isn't refetched automatically, so you only see it load when navigating to the page. Which to me is an extremely reasonable and expected thing when loading a page. The query is ~~invalidated~~ set to stale automatically thanks to the QueryCache staleTime config of 30 seconds, but again, that does not automatically refetch.

src/routers/v1/artists/{id}/index.ts
timglorioso: 1. Yes
2. No
3. Will do

timglorioso: Here's what the singleInclude adds. I could modify this to include full artistLabel info.

What's the purpose of singleInclude? ~~It only works for including things on Artist?~~ oh it's artist utils. Hm. Still don't quite understand what it means to single include.

https://github.com/funmusicplace/mirlo/blob/5d8322b477d55a954ce41d53054a988688fa2bad/src/utils/artist.ts#L578-L591

simonv3: It's everything to include when returning a single artist. It's not super optimized though, and could probably use some attention. But it's basically like "the artist page depends on these things".

timglorioso: Cool, thanks, I think it makes sense to return full artistLabel info here in the Artist as label case. That's the intent here: to use the /artists/{artistId} endpoint to get the artist profile used by the label. It's not ideal, but it was the best path I found to fix this issue.

client/src/components/Label/Roster.tsx
simonv3: Hmm, I think I disagree on the loading stuff... I actually just spent a lot of time making the artist page load pretty snappy with data injection happening into the header scripts so that there isn't a loading spinner at all on the artist page when you navigate to it directly. This view specifically reloading on isFetching actually bothered me significantly when I would return obsessively to a certain label page from another tab and it updated the entire page, all artists disappearing and then reloading. 

I like what [it says here](https://medium.com/@manoj.munda/tanstack-query-isloading-ispending-and-isfetching-89db5f45d7d9). 

> The isFetching flag is trueany time a fetch is in progress, whether it's the initial load or a background refetch. This makes it especially useful for showing subtle loading indicators while data is being refreshed in the background, without disrupting the current UI.
> 
> Use case: Suppose you have a list component on your page that updates in the background. In that case, instead of a full loading screen, you can just show a small loader right inside that list area while it fetches new data.

A subtle loading indicator rather than hiding the entire screen might be a better bet? I did just change [this last month](https://github.com/funmusicplace/mirlo/commit/cfe21a35d094cb451cd109e074bf442303e26cbc#diff-d72d04565f3f60ee503f835d57e4244944207bfc6c23e2348f47cf828ffcbab7R22-L36) haha.

timglorioso: Thanks for the context and feedback. I was so focused on the delete flow, I didn't realize the impact elsewhere. Reverting to isPending.

PR #2040, opened 2026-05-06
Fix: Form saves cleanup

LouisLouiseKay:
This is a follow up to #1950 and #1977 to fix the issue with autosave where any change on a publicly visible album or merch or blogpost would go live immediately on autosave input. This removes all autosave inputs from merch, blogposts, and album release pages, and replaces them with browser local storage with a "keep / discard" banner when a user comes back to a release with unsaved changes. While local storage isn't perfect by any means (doesn't handle multi-device editing, doesn't handle conflicts etc.) I think it'll be an improvement still and will do as expected in the vast majority of use cases. Ultimately we'd need to build a proper draft structure on the backend but that was out of my range.


client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PaymentSlider.tsx
```diff
@@ -1,37 +1,30 @@
 import { css } from "@emotion/css";
```
simonv3, line 1: This component gets used across some other places so just checking that they still work there.

LouisLouiseKay: Yep, taken into account it's left untouched anywhere else it's used.

client/src/components/ManageArtist/Posts/PostForm.tsx
```diff
@@ -87,90 +87,60 @@ const PostForm: React.FC<{
...
 
   const isPublic = watch("isPublic");
-  const minimumTier = watch("minimumTier");
 
   const existingId = post.id;
-  const userId = user?.id;
 
   const publicationDate = watch("publishedAt");
 
-  const doSave = React.useCallback(
```
simonv3, line 101: Wait so how is this saved?

LouisLouiseKay: Ah this is actually dead code, the save was never really in PostForm. I remove it in the next commit (I should have switched order? I wasn't sure). Now that there's no autosave input it's all moved to the "Save draft" / "Update post" button and PublishPostButton. They use the form values with useFormContext and PUTs them to "manage/posts/:id" (and the Publish button also PUTS to /publish obviously). PostForm just has form state, fields, and the delete button. SaveDraft and Publish also activate onSaveSuccess after the PUT has worked so that we clear the draft that's still in localStorage.

simonv3 (dismissed):

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/saveTrackGroupForm.ts
```diff
@@ -0,0 +1,111 @@
...
+    const values = methods.getValues();
+    try {
+      await saveTrackGroupForm(values, {
+        trackGroupId: trackGroup.id,
+        artistId,
+        fundraiserId: trackGroup.fundraiser?.id,
+      });
+      methods.reset(values);
+      await reload();
+      client.invalidateQueries({
```
simonv3: If we're clearing caches here, do we still need the `reload` stuff? 

Likewise I feel like this could be a bit cleaner if `saveTrackGroupForm` could be useQuery mutation.

client/src/components/ManageArtist/Posts/PostForm.tsx
```diff
@@ -52,29 +47,77 @@ const PostForm: React.FC<{
...
-        }
-      : {
-          publishedAt: publishedAt.toISOString().slice(0, 16),
-          shouldSendEmail: true,
-          isPublic: true,
-        },
+    defaultValues: buildDefaultValues(),
   });
 
+  const formDraftKey = post?.id ? `postDraft-${post.id}` : null;
```
simonv3, line 84: I feel like all of this could use some explanation on why it's separated and slightly verbose?

client/src/utils/useBodyDraft.ts
```diff
@@ -0,0 +1,82 @@
+import React from "react";
```
simonv3, line 1: Could you add some comments here on why we have a specific post draft handler?

client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/PriceAndSuch.tsx
```diff
@@ -332,24 +261,87 @@ const PriceAndSuch: React.FC<{
...
+          `}
+        >
+          <FormComponent
             className={css`
-              width: 100%;
-              @media screen and (min-width: ${bp.medium}px) {
-                display: grid;
-                grid-template-columns: repeat(2, 1fr);
-              }
+              flex-grow: 1;
```
simonv3: Tailwind pls

client/src/components/ManageArtist/Posts/PostForm.tsx
LouisLouiseKay: Agreed, I'll add a few words, working around remirror's limitations wasn't straightforward _at all_ (we might want to investigate some other rich text editor in the future I think)

client/src/utils/useBodyDraft.ts
LouisLouiseKay: Yeah, added that as a follow up to the post body content comment

simonv3 (approved):

PR #2080, opened 2026-05-13
feat: use JWT playtoken for song streaming

txels:
Starting on the Mirlo side of authorized HLS playback as specced in https://codeberg.org/fairplayer/interop/pulls/10.

- Mirlo generates a playback token upon first request for a song from a user (the playlist.m3u8 endpoint) based on headers provided by the client (this means Fairplayer for now)
- For segments, it bases the authorization on the playtoken and userid passed in as a querystring

This doesn't solve CORS for segments though, for which we may need some extra work.

The code could be made cleaner.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,10 +1,14 @@
+approvedGitRepositories:
+  - "**"
+
 compressionLevel: mixed
 
-nodeLinker: node-modules
+enableScripts: true
 
 logFilters:
-  # Ignores some known warnings about the react/react-dom versions not satisfying @gumlet/react-hls-player
-  - pattern: "*(paf38f)*"
-    level: discard
-  - pattern: "*(p741fc)*"
-    level: discard
+  - level: discard
+    pattern: "*(paf38f)*"
+  - level: discard
+    pattern: "*(p741fc)*"
+
+nodeLinker: node-modules
--- a/README.md
+++ b/README.md
@@ -15,12 +15,12 @@ Prerequisites:
 
 - [Docker](https://www.docker.com/get-started/)
 - [Node.js 22](https://nodejs.org/en)
-- [Yarn 4](https://yarnpkg.com)
 
 Once node is installed, you can use [Corepack](https://nodejs.org/api/corepack.html) to automatically install the correct version of yarn (as defined in `package.json`):
 
 ```sh
 corepack enable
+corepack up
 ```
 
 If you'd like to use Nix, we have a basic `shell.nix` that includes Node.js, Yarn, and other tools to develop the client. Activate the shell using `nix-shell` at the top-level of the repository.
--- a/docker-compose.api.yml
+++ b/docker-compose.api.yml
@@ -27,6 +27,7 @@ services:
     environment:
       - NODE_ENV=${NODE_ENV:-development}
       - REDIS_HOST=${REDIS_HOST:-redis}
+      - DATABASE_URL=postgresql://nomads:nomads@pgsql:5432/nomads?schema=public
     depends_on:
       - redis
       - pgsql
--- a/package.json
+++ b/package.json
@@ -4,10 +4,11 @@
   "license": "AGPL-3.0",
   "module": "commonjs",
   "scripts": {
+    "api:debug": "dotenv ts-node src/index.ts  --inspect",
+    "api:start": "ts-node src/index.ts",
     "api:test": "tsc --noEmit && mocha test --exit --extension ts",
     "api:test:docker": "docker compose -f docker-compose.api.yml -f docker-compose.test.yml --profile test run --rm test",
-    "api:watch": "NODE_ENV=development PORT=3000 nodemon src/index.ts",
-    "api:start": "ts-node src/index.ts",
+    "api:watch": "NODE_ENV=development PORT=${PORT:-3000} nodemon src/index.ts",
     "prepare": "husky",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
@@ -121,6 +122,7 @@
     "@types/yargs": "^17.0.22",
     "@yarnpkg/pnpify": "^4.0.1",
     "concurrently": "^7.6.0",
+    "dotenv-cli": "^11.0.0",
     "lint-staged": "^15",
     "mocha": "^10.2.0",
     "nodemon": "^2.0.20",
@@ -148,5 +150,5 @@
     "prisma",
     "client"
   ],
-  "packageManager": "yarn@4.1.1"
+  "packageManager": "yarn@4.14.1+sha512.64df448055b2d37ba269d7db535a469b8da93f8ef1140c25fd7a83c00a8fbaacb214ca0e02553b92a2c54cef78bb67d0b4817fab02001df0e24fac0faccc3b42"
 }
--- a/src/auth/cors.ts
+++ b/src/auth/cors.ts
@@ -43,6 +43,7 @@ export const corsCheck = async (...args: [Request, Response, NextFunction]) => {
   // @ts-ignore - req.logger added by middleware
   const log = req.logger || logger;
   try {
+    const isNotCors = req.headers["sec-fetch-mode"] === "no-cors";
     const isSameSite =
       req.headers["sec-fetch-site"] === "same-site" ||
       req.headers["sec-fetch-site"] === "same-origin";
@@ -69,15 +70,27 @@ export const corsCheck = async (...args: [Request, Response, NextFunction]) => {
         req.method === "OPTIONS" &&
         req.headers["access-control-request-method"] &&
         req.headers["access-control-request-headers"];
-      const skipsApiKey =
+      let skipsApiKey =
+        isNotCors ||
         isSameSite ||
         !isAPIEndpointPrivate ||
         isCORSPreflight ||
         isValidActivityPubEndpoint(req.path);
+      const origin = req.get("origin");
 
       log.info(
-        `CORS check for ${req.method} ${req.path} - sameSite: ${isSameSite}, privateEndpoint: ${isAPIEndpointPrivate}, skipsApiKey: ${skipsApiKey}, isCorsPreflight: ${isCORSPreflight}`
+        `CORS check for ${origin} ${req.method} ${req.path} - sameSite: ${isSameSite}, privateEndpoint: ${isAPIEndpointPrivate}, skipsApiKey: ${skipsApiKey}, isCorsPreflight: ${isCORSPreflight}`
       );
+
+      if (origin) {
+        const clientsThatMaySkipToken = await prisma.client.findMany({
+          where: { allowedCorsOrigins: { has: origin }, skipStreamToken: true },
+        });
+        skipsApiKey ||= clientsThatMaySkipToken.length > 0;
+        log.info(
+          `Found clients that don't require token: ${clientsThatMaySkipToken.length}`
+        );
+      }
 
       log.info(
         "headers",
--- a/src/auth/cors.ts
+++ b/src/auth/cors.ts
@@ -75,7 +75,20 @@ export const corsCheck = async (...args: [Request, Response, NextFunction]) => {
         isCORSPreflight ||
         isValidActivityPubEndpoint(req.path);
 
+      log.info(
+        `CORS check for ${req.method} ${req.path} - sameSite: ${isSameSite}, privateEndpoint: ${isAPIEndpointPrivate}, skipsApiKey: ${skipsApiKey}, isCorsPreflight: ${isCORSPreflight}`
+      );
+
+      log.info(
+        "headers",
+        req.headers["access-control-request-method"],
+        req.headers["access-control-request-headers"]
+      );
+
       if (skipsApiKey) {
+        log.info(
+          `${req.method} ${req.path} Skipping API key check for ${req.method} ${req.path}`
+        );
         clients = await prisma.client.findMany();
       } else {
         const apiHeader = req.headers[MIRLO_API_KEY_HEADER];
--- a/src/routers/v1/tracks/{id}/stream/authzd/{segment}.ts
+++ b/src/routers/v1/tracks/{id}/stream/authzd/{segment}.ts
@@ -1,0 +1,169 @@
+import prisma from "@mirlo/prisma";
+import { NextFunction, Request, Response } from "express";
+import jwt from "jsonwebtoken";
+
+import { userLoggedInWithoutRedirect } from "../../../../../../auth/passport";
+import logger from "../../../../../../logger";
+import {
+  finalAudioBucket,
+  getBufferBasedOnStat,
+  statFile,
+} from "../../../../../../utils/minio";
+import { canUserListenToTrack } from "../../../../../../utils/ownership";
+import socialMusic from "../../../../../../utils/socialMusic";
+
+const jwt_secret = process.env.JWT_SECRET ?? "secretkey";
+
+export const fetchFile = async (
+  res: Response,
+  filename: string,
+  segment: string
+) => {
+  const alias = `${filename}/${segment}`;
+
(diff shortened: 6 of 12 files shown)


.yarnrc.yml
```diff
@@ -1,10 +1,14 @@
+approvedGitRepositories:
+  - "**"
+
```
txels, line 3: This was added by tooling, I presume...

package.json
```diff
@@ -148,5 +150,5 @@
     "prisma",
     "client"
   ],
-  "packageManager": "yarn@4.1.1"
+  "packageManager": "yarn@4.14.1+sha512.64df448055b2d37ba269d7db535a469b8da93f8ef1140c25fd7a83c00a8fbaacb214ca0e02553b92a2c54cef78bb67d0b4817fab02001df0e24fac0faccc3b42"
```
txels, line 153: Not sure why this happened TBH, can revert.

src/routers/v1/tracks/{id}/stream/{segment}.ts
```diff
@@ -69,14 +72,45 @@ export default function () {
...
+            const playToken = jwt.sign(payload, jwt_secret, {
+              expiresIn: "4w",
+            });
+            res.setHeader("x-playtoken", playToken);
+          }
+        }
+
+        const playToken = req.query.playtoken;
+        const reqUserId = req.query.userid;
+        if (!isManifest && playToken) {
```
txels: We need to figure out how to enforce the token in the case of Fairplayer requests for segments that won't include custom headers. In other words, how to tell apart requests from legit Mirlo web client and those from a federated client, which should all be CORS-allowed without custom HTTP headers, but the federated clients will be required to include the playtoken.

WDYT @simonv3 ?

src/routers/v1/tracks/{id}/stream/{segment}.ts
```diff
@@ -69,14 +72,45 @@ export default function () {
...
-        segment.includes("playlist.m3u8") &&
-        isUserAbleToListenToTrack === "exceeded"
-      ) {
+      const isManifest = segment.includes("playlist.m3u8");
+      if (isManifest && isUserAbleToListenToTrack === "exceeded") {
         res.status(402).send("Track play limit exceeded");
       }
 
       if (track.audio) {
+        // TODO: come up with the right header names for the "social music network"
```
simonv3: `social-music-userid`?

src/routers/v1/tracks/{id}/stream/{segment}.ts
simonv3: I was thinking about checking `req.origin` but the issue is that our mobile app should also pass through this unscathed (for now, until we set it up to work the same?). Our ability to support the mobile app is a bit slower so it would take a while to get something new up and deployed and rolled out. What might make sense is to have a boolean on the client table to let them skip, and then in our client check (which, I know, is in the to-be-refactored-all-in-one-CORS-middleware) set something on the request that's like `req.skipStreamToken` and use that for now? 

what do you think?

simonv3: Wondering if it makes sense to just have a second route that does something like this.

```
src/routers/v1/tracks/{id}/stream/remote/{segment}.ts
...    
GET: [userLoggedInWithoutRedirect, checkJWTPlayToken, GET],
...
```

Where the original GET is the same in both files? Has the benefit of being backwards compatible and has the ability to continue doing custom things in it. Could also instead of `remote` do something that's more basic like `routers/v2/tracks/{id}...`

txels: Ah maybe a v2 makes sense.
But then the content syndication endpoint (aka "canimus") should give the correct URLs for the m3u8 file within the v2 API. The segment parent paths have to match the manifest's because the manifest contains relative URLs and you don't want to recompute those.

src/routers/v1/tracks/{id}/stream/{segment}.ts
txels: I can't remember now whether the `x-` prefix is in or out of favor.

src/routers/v1/tracks/{id}/stream/{segment}.ts
simonv3: Yeah I think that'd be fine (cc @MaciAC)

src/routers/v1/tracks/{id}/stream/authzd/{segment}.ts
```diff
@@ -0,0 +1,166 @@
...
+    return;
+  }
+};
+
+export default function () {
+  const operations = {
+    GET: [userLoggedInWithoutRedirect, GET],
+  };
+
+  async function GET(req: Request, res: Response, next: NextFunction) {
```
txels, line 18: So this is now on a separate route as recommended by @simonv3 . I struggled too much to try to make it work on a `/v2` path and gave up, it seemed to require more work than just adding the file to the path, and how everything was tied up with the swagger docs... it felt like too much trouble TBH.

src/routers/v1/tracks/{id}/stream/{segment}.ts
txels: Struggled too much (for what I considered a reasonable amount of time) with setting a v2 path, so I put it elsewhere in a `stream/authzd` path.

txels: > GET: [userLoggedInWithoutRedirect, checkJWTPlayToken, GET],

I haven't done this yet (the check token as a middleware). Not sure how much worth the effort it is considering the playtoken query string argument is only used in this one endpoint, that the endpoint is used for both manifest and segment, and both have different validation mechanics. I've now refactored the code to show the clear split logic paths for manifest and segment.

src/routers/v1/tracks/{id}/stream/authzd/{segment}.ts
txels: Making it a separate route has made the check logic simpler to express so that is a clear win.

src/routers/v1/tracks/{id}/stream/authzd/{segment}.ts
```diff
@@ -0,0 +1,166 @@
...
+      );
+
+      if (!track || !isUserAbleToListenToTrack) {
+        res.status(404);
+        return next();
+      }
+
+      const isManifest = segment.includes("playlist.m3u8");
+      if (isManifest && isUserAbleToListenToTrack === "exceeded") {
+        res.status(402).send("Track play limit exceeded");
```
txels, line 50: I guess I should return next() after every error status? I'm too new to vanilla express.

.yarnrc.yml
txels: I've retraced my steps, removed these changes, called `corepack enable && corepack up` and this showed up again.

Is that the correct way to get started or not?

tsconfig.json
```diff
@@ -8,7 +8,10 @@
     "skipLibCheck": true,
     "esModuleInterop": true,
     "moduleResolution": "node16",
-    "isolatedModules": true
+    "isolatedModules": true,
+    "paths": {
+      "@/*": ["./src/*"]
```
txels: Failed attempt at being able to import `"@utils/..."` rather than `../../../../utils`. The IDE likes it but the command line doesn't.

.yarnrc.yml
timglorioso: I've never called `corepack up` FWIW

src/routers/v1/tracks/{id}/stream/authzd/{segment}.ts
timglorioso: I'm no Express expert either, but I believe `res.send` ends the request-response cycle and does not need to call `next`

https://expressjs.com/en/guide/writing-middleware.html

txels: We still need to return somehow, not to continue processing the request in the function...

src/auth/cors.ts
```diff
@@ -43,6 +43,7 @@ export const corsCheck = async (...args: [Request, Response, NextFunction]) => {
   // @ts-ignore - req.logger added by middleware
   const log = req.logger || logger;
   try {
+    const isNotCors = req.headers["sec-fetch-mode"] === "no-cors";
```
txels, line 46: I noticed that the segment URLs as loaded by the player don't really include an origin header, but they do include this one.

src/auth/cors.ts
```diff
@@ -69,13 +70,38 @@ export const corsCheck = async (...args: [Request, Response, NextFunction]) => {
...
         req.headers["access-control-request-method"] &&
         req.headers["access-control-request-headers"];
-      const skipsApiKey =
+      let skipsApiKey =
+        isNotCors ||
         isSameSite ||
         !isAPIEndpointPrivate ||
         isCORSPreflight ||
         isValidActivityPubEndpoint(req.path);
+      const origin = req.get("origin");
```
txels, line 79: This origin header I've sent manually from fairplayer when doing the authed request with an API, where I actually include the token, so we do not really need this.

src/auth/cors.ts
```diff
@@ -69,13 +70,38 @@ export const corsCheck = async (...args: [Request, Response, NextFunction]) => {
...
+
+      if (origin) {
+        const clientsThatMaySkipToken = await prisma.client.findMany({
+          where: { allowedCorsOrigins: { has: origin }, skipStreamToken: true },
+        });
+        skipsApiKey ||= clientsThatMaySkipToken.length > 0;
+        log.info(
+          `Found clients that don't require token: ${clientsThatMaySkipToken.length}`
+        );
+      }
```
txels, line 93: this whole block is not needed for the reason mentioned above

src/auth/cors.ts
txels: Ah wait, this is different on firefox than chromium... on FF the origin is included and the request is "cors".

src/auth/cors.ts
txels: ...well it is needed on Firefox

src/routers/v1/tracks/{id}/stream/authzd/{segment}.ts
timglorioso: Discussion here suggests to simply `return` after `res.send`

https://stackoverflow.com/questions/14709802/exit-after-res-send-in-express-js?noredirect=1&lq=1

README.md
```diff
@@ -15,12 +15,12 @@ Prerequisites:
 
 - [Docker](https://www.docker.com/get-started/)
 - [Node.js 22](https://nodejs.org/en)
-- [Yarn 4](https://yarnpkg.com)
```
timglorioso, line 18: Is this because you technically don't install Yarn directly?

txels: It came as part of the `corepack` setup. Partly this is also an attempt from me to get reactions as to what should be the "friction-free setup for newbies".

src/routers/v1/tracks/{id}/stream/authzd/{segment}.ts
```diff
@@ -0,0 +1,172 @@
...
+  finalAudioBucket,
+  getBufferBasedOnStat,
+  statFile,
+} from "../../../../../../utils/minio";
+import { canUserListenToTrack } from "../../../../../../utils/ownership";
+import socialMusic from "../../../../../../utils/socialMusic";
+
+const jwt_secret = process.env.JWT_SECRET ?? "secretkey";
+
+export const fetchFile = async (
```
simonv3: Can this be re-used from the other file?

src/routers/v1/tracks/{id}/stream/authzd/{segment}.ts
simonv3: Yeah its own route is fine by me. It also matches how we've done separate routes for AP in other places.

src/routers/v1/tracks/{id}/stream/authzd/{segment}.ts
txels: Yes of course, should we move it into a library path outside `routers` then? What would your preferred location be?

simonv3: imo it's fine to just import it from that endpoint. But maybe it fits in the `utils/minio.ts` library?

txels: I've imported from the other endpoint as I've seen this was already done elsewhere and moving it to utils would have required making the function slightly more generic (it would feel weird that the utils library new about 'segments').

src/auth/cors.ts
txels: So the firefox vs brave differences from what I could gather are due to Brave supporting the m3u8 file natively in an audio tag and issuing non-CORS requests vs Firefox not supporting it and requiring hls.js (which does issue Ajax requests that are tagged as cross-origin).

On that note my suggested follow up re: CORS is:
- split the current cors middleware into two, separating cors handling from API key/auth requirements
- in the new cors middleware, only decide for which requests to return CORS headers based on routes (some routes always support CORS) and origin (we allow CORS for some origins/clients - we could change this client flag to be called `allowCors` rather that the `skipStreamToken` I've added in the current version of this PR).
- in the API key middleware, we block access to certain endpoints if no key is present. This middleware we could decide to explicitly add it only to routes that require an API key, or add it "globally" and let it know what routes are "private" (which is how it's done now, so we could keep that for simplicity). Maybe there are better ways in express to have two "sets of routes" and apply certain middleware to only one of those.

src/routers/v1/tracks/{id}/stream/{segment}.ts
simonv3: Yeah I don't think it's worth the effort.

src/auth/cors.ts
simonv3: That sounds good to me @txels do you mind giving it a shot?

docker-compose.api.yml
```diff
@@ -27,6 +27,7 @@ services:
     environment:
       - NODE_ENV=${NODE_ENV:-development}
       - REDIS_HOST=${REDIS_HOST:-redis}
+      - DATABASE_URL=postgresql://nomads:nomads@pgsql:5432/nomads?schema=public
```
simonv3, line 30: Why was this needed? 

Either way, if you do this if the user sets a DATABASE_URL in an env variable it'll pick that up and default to the other thing if not. 
```suggestion
      - DATABASE_URL=${DATABASE_URL:postgresql://nomads:nomads@pgsql:5432/nomads?schema=public}
```

package.json
simonv3: Yeah maybe remove this? Though maybe that's the new way that yarn does version management.

src/auth/cors.ts
txels: Happy to give it a shot, however some pairing/guidance on setting up tests or enumerating test cases for CORS and auth would help, as I don't have enough context and I don't think there is currently unit test coverage for those, or at least haven't found them.

docker-compose.api.yml
txels: Ah I can revert, this was me trying to use the same .env file for running things outside Docker and inside it. Let me just revert all setup changes and leave only the actual code changes in the PR.

src/auth/cors.ts
simonv3: I'll put in something later today I hope!

simonv3: https://github.com/funmusicplace/mirlo/pull/2120/changes

txels: Awesome, I will recreate this PR (the good bits) on top of that branch.

simonv3: merged it! Hope it doesn't break anything.

txels:
Closing in favour of #2128

src/auth/cors.ts
txels: All good! Thanks

PR #2088, opened 2026-05-14
fix: restore label roster artists avatars, backgrounds and cover mosaics

LouisLouiseKay:
This is a quick fix for a bug that I realized while testing locally but is already active on prod so I prioritized it immediately because artist avatars don't appear on label pages anymore.

The cause is a change in architecture, where /v1/artists/{slug} endpoint is now returning user.artistLabels but just the artist scalar fields, so no avatar, background and trackgroup. The label roster page (which now reuses queryArtist instead of the old specific endpoint) was rendering an empty placeholder for every artist instead of their actual avatar, or the cover mosaic fallback for artists without an avatar, or the background image.

This adds back the avatar, background and most recent published trackgroups includes on the nested artist.

### DIFF
--- a/src/utils/artist.ts
+++ b/src/utils/artist.ts
@@ -27,7 +27,11 @@ import {
 } from "./minio";
 import { getSiteSettings } from "./settings";
 import stripe from "./stripe";
-import { deleteTrackGroup, trackGroupPublishedObject } from "./trackGroup";
+import {
+  deleteTrackGroup,
+  trackGroupPublishedObject,
+  whereForPublishedTrackGroups,
+} from "./trackGroup";
 export { processSingleArtist } from "./serialize/artist";
 
 type Params = {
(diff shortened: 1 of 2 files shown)


src/utils/artist.ts
```diff
@@ -594,7 +594,21 @@ export const singleInclude = (queryOptions?: {
...
+                  where: trackGroupPublishedObject(),
+                  include: {
+                    cover: true,
+                    tracks: true,
+                  },
+                  orderBy: { releaseDate: "desc" },
+                  take: 4,
+                },
+              },
+            },
```
timglorioso, line 619: 🙏 
For my knowledge, why different from what the endpoint used to be? https://github.com/funmusicplace/mirlo/blob/5d16cabbd4086fdae51151409ef7e62c32fdd855/src/routers/v1/labels/%7Bid%7D/index.ts#L43-L64

timglorioso:
I pulled this down and it's working for me

src/utils/artist.ts
LouisLouiseKay: Oh some of that might actually be good to add, I really *just* added back the stuff that went missing in a hurry and checked that it worked on my end. whereForPublishedTrackGroups especially might actually be better cause I *just* added the full trackgroups which could cause issues. The only "important" thing is "take: 4" so we only take for the four last releases, for the payload stuff cause now with the new system if I understand correctly we'd otherwise be... fetching ALL the trackgroup covers of ALL the artists of a label right?

LouisLouiseKay: (The "Four releases" is for the mosaic stuff @simonv3 added in case an artist doesn't have Avatar by the way)

timglorioso: Gotcha yeah `take: 4` makes sense for the placeholder image

LouisLouiseKay: OK just copied the old code and added the four covers limitation.

simonv3 (approved):

PR #2093, opened 2026-05-15
Declutter artist pages

LouisLouiseKay:
There was a lot of visual noise on the artist pages related to editing EVEN when in "view mode", which meant an artist could never really see their page "as listeners see it". The problem was particularly pronounced on mobile, with lots of duplicated icons, sometimes even going as far as breaking the visual layout on smaller smartphones.

### DIFF
--- a/client/src/components/Artist/Artist.tsx
+++ b/client/src/components/Artist/Artist.tsx
@@ -8,7 +8,7 @@ import {
 } from "queries";
 import React from "react";
 import { useTranslation } from "react-i18next";
-import { FaChevronRight, FaEdit } from "react-icons/fa";
+import { FaChevronRight } from "react-icons/fa";
 import {
   NavLink,
   Outlet,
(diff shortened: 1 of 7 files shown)


timglorioso:
Nice, when I added all the labeling I was tempted to just remove the buttons in the first place, but didn't want to rock the boat too much heh.

LouisLouiseKay:
Yeah it was a mess, partial leftover of initial debates where we didn't make a clear cut choice between a wysiwyg approach (same page with edit buttons everywhere in edit mode / clean page in view mode) and a more "admin template" approach (table views, more affordance and laid out to display as much info as possible), I think we're slowly moving towards fully admin based view on the manage side (and we can still make another run at a wysiwyg approach later and be more mindful about it), which means this view is obsolete and people should be able to see as close as a 1:1 page in view mode as possible.

simonv3 (approved):

PR #2098, opened 2026-05-15
Player overhaul

LouisLouiseKay:
Player was using old stylings (merged to common component logic with Embeds now), that would most often extend its height and take up lots of screen real estate while also making track titles illegible in a lot of cases. Similarly, buttons over the player didn't display correctly, and had incoherent non matching styles. The "x plays left" indicator was also clashing with the progress bar and overflowing on those buttons. The player now also uses the same system as the header bar to adjust to the artist theme (very simple flip/flop between light or dark depending on artist's background color). This also fixes a bug where the player wouldn't play when hit for the first time on Mirlo for someone that had never been on the website before, and another bug where clicking "previous" would make the player disappear despite being on an album page. 

Also added a cypress test to make sure the player doesn't remount upon browsing the website and continuity is always garanteed.

### DIFF
--- a/client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/FundraisingGoal.tsx
+++ b/client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/FundraisingGoal.tsx
@@ -132,10 +132,18 @@ const FundraisingGoal: React.FC<{
 
   if (!fundraiser) {
     return (
-      <div>
+      <div className="flex flex-col items-start">
         <h2>{t("addFundraiser")}</h2>
-        <p>{t("addFundraiserDescription")}</p>
-        <Button type="button" onClick={onAddFundraiser} isLoading={isLoading}>
+        <small id="description-add-fundraiser">
+          {t("addFundraiserDescription")}
+        </small>
+        <Button
+          type="button"
+          onClick={onAddFundraiser}
+          isLoading={isLoading}
+          aria-describedby="description-add-fundraiser"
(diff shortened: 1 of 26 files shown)


client/src/components/ManageArtist/ManageTrackGroup/AlbumFormComponents/FundraisingGoal.tsx
```diff
@@ -132,10 +132,18 @@ const FundraisingGoal: React.FC<{
...
     return (
-      <div>
+      <div className="flex flex-col items-start">
         <h2>{t("addFundraiser")}</h2>
-        <p>{t("addFundraiserDescription")}</p>
-        <Button type="button" onClick={onAddFundraiser} isLoading={isLoading}>
+        <small id="description-add-fundraiser">
+          {t("addFundraiserDescription")}
+        </small>
+        <Button
```
simonv3, line 140: ArtistButton?

client/src/components/Player/index.tsx
```diff
@@ -68,184 +82,198 @@ const Player = () => {
     return null;
   }
 
-  return (
+  return createPortal(
```
simonv3, line 85: Can you document why we need this?

simonv3 (approved):

client/src/components/Player/index.tsx
LouisLouiseKay: Yeah I'll add a comment in a "small fixes" PR I'm on

PR #2103, opened 2026-05-17
fix: clearer error and friendlier UX when downloading a release that …

### LINKED ISSUE #1773: Download failing shortly after purchase
Purchase completed Feb 18 12:42:26 PM, download attempted Feb 18 12:42:50 PM. Despite the error "no purchase found", whether or not the download failing has anything to do with time since purchase is unknown. Download eventually worked a few hours later.

> After purchase, I go to the album click the download link, it then 'generates
> the download' and then shows a download button. When I click that it
> errors to what looks like a json string saying no purchase was found -
> the album is never downloaded.
>
> error "trackGroupId: 3730 no purchase found"
>
> I can see in the Admin panel that the purchase has been completed and a
> sales record is generated.

> Seeing a couple of these in the logs error: Error fetching static file from backblaze: trackgroup-format/3730/wav.zip {"service":"api"} which is probably related. Feels like maybe there's some front-end request happening before it's quite finished with processing, which maybe we need to handle better.

peterklingelhofer:
…isn't ready yet

Closes #1773

### DIFF
--- a/client/src/components/common/DownloadAlbumButton.tsx
+++ b/client/src/components/common/DownloadAlbumButton.tsx
@@ -1,15 +1,14 @@
+import { ArtistButton } from "components/Artist/ArtistButtons";
 import React from "react";
-import { css } from "@emotion/css";
 import { useTranslation } from "react-i18next";
 import { RiDownloadLine } from "react-icons/ri";
 import api from "services/api";
-import { useSnackbar } from "state/SnackbarContext";
-import { useArtistContext } from "state/ArtistContext";
-import Modal from "./Modal";
-import LoadingSpinner from "./LoadingSpinner";
-import { ArtistButton } from "components/Artist/ArtistButtons";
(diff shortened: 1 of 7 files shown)


client/src/components/common/DownloadAlbumButton.tsx
```diff
@@ -53,11 +51,72 @@ const DownloadAlbumButton: React.FC<{
...
       queryParams.append("token", token);
     }
+    const endpoint = `${prefix}/download?${queryParams.toString()}`;
 
-    return api.getFileDownloadUrl(
-      `${prefix}/download?${queryParams.toString()}`
-    );
-  }, [track, trackGroup.id, prefix, chosenFormat, email, token]);
+    setIsDownloading(true);
+    try {
```
simonv3, line 62: I'm wondering if this should actually just be in `getFileDownloadUrl`?

simonv3 (changes requested):

PR #2107, opened 2026-05-17
feat: let users reset a disconnected Stripe account with email + pass…

### LINKED ISSUE #2085: reset stripe account
> I finished setting up Stripe, and only then realised that I had accidentally created a new Stripe account, rather than connecting Mirlo to my existing Stripe account. I deleted the newly created Stripe account, and then went back to Mirlo to attempt to re-start the process. However, it appears that Mirlo’s backend does not account for a user deleting their Stripe account; so when I press ‘Set up bank account’ it takes me to a page displaying this raw-JSON Stripe API response:
> 
> {"error":"Unable to access Stripe account. The API key may not have permission to access this account or the account may have been deleted."}
> 
> If there’s any way to erase the database record referencing my deleted Stripe account so that I may start afresh, then that would be much appreciated.

We should make a button on the user's account page that says "reset'. When they do that we should make them verify their password and their e-mail address.

peterklingelhofer:
…word confirmation

Closes #2085


client/src/components/common/stripe/StripeStatusAndButton.tsx
```diff
@@ -4,22 +4,151 @@ import api from "services/api";
...
+// stale id, every subsequent "Set up bank account" attempt bounces off Stripe
+// with an opaque error. See #2085.
+const ResetStripeAccountModal: React.FC<{
+  userId: number;
+  userEmail: string;
+  onReset: () => void;
+}> = ({ userId, userEmail, onReset }) => {
+  const { t } = useTranslation("translation", { keyPrefix: "manage" });
+  const snackbar = useSnackbar();
+  const [open, setOpen] = React.useState(false);
```
simonv3: Use react hook form maybe?

simonv3:
This is great, but I realized I should have added more details to the ticket--when I meant "verify their e-mail address" I meant a kind of 2fa code to be sent to them (similar to how we do elsewhere in the UI) so that we verify the user actually has address to the e-mail address.

simonv3 (changes requested):

simonv3 (approved):

PR #2108, opened 2026-05-17
feat: mark fundraisers SUCCESSFUL after charge and switch trackGroup …

### LINKED ISSUE #1681: Marking a fundraiser as complete and making it so that now people can just buy the release instead of pledging.
Probably have to store this as a boolean on the track group, just adding `fundraisingComplete`.

peterklingelhofer:
…to regular purchase

Closes #1681


simonv3 (commented):
This looks good but I think it's missing a button that lets the artist say "this fundraiser is done!" This would just call the `chargePledges` endpoint. 

Sorry that could have been scoped better in the ticket.

simonv3 (changes requested):

simonv3 (approved):
My bad there's already a button it just didn't always appear

PR #2109, opened 2026-05-17
feat: embed Stripe subscription checkout in-app instead of redirectin…

### LINKED ISSUE #1168: Stripe checkout should go into a pop up window
Absolute buzz kill that you're listening to a song and then you press buy and the song stops.

peterklingelhofer:
…g to Stripe

Closes #1168

### DIFF
--- a/src/routers/v1/artists/{id}/subscribe.ts
+++ b/src/routers/v1/artists/{id}/subscribe.ts
@@ -126,10 +126,8 @@ export default function () {
       });
       logger.info(`Generated a Stripe checkout session ${session.id}`);
 
-      // Embedded mode returns a client_secret instead of a session url; the
-      // client renders <EmbeddedStripeForm> inline rather than redirecting
-      // to Stripe's hosted page (#1168).
       res.status(200).json({
+        sessionUrl: session.url,
         clientSecret: session.client_secret,
         stripeAccountId,
       });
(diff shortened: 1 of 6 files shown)


src/routers/v1/artists/{id}/subscribe.ts
```diff
@@ -127,8 +127,12 @@ export default function () {
       });
       logger.info(`Generated a Stripe checkout session ${session.id}`);
 
+      // Embedded mode returns a client_secret instead of a session url; the
+      // client renders <EmbeddedStripeForm> inline rather than redirecting
+      // to Stripe's hosted page (#1168).
       res.status(200).json({
-        sessionUrl: session.url,
```
simonv3, line 130: I think we want to keep sessionUrl for redirects from _outside_ of Mirlo!

test/routers/artists/{id}.subscribe.spec.ts
```diff
@@ -92,9 +92,23 @@ describe("artists/{id}/subscribe", () => {
...
+      assert.ok(
+        "clientSecret" in body,
+        `expected clientSecret key in response, got: ${response.text}`
+      );
+      assert.ok(
+        typeof body.stripeAccountId === "string" &&
+          body.stripeAccountId.length > 0,
+        `expected stripeAccountId in response, got: ${response.text}`
+      );
+      assert.ok(
```
simonv3: Note above

simonv3 (changes requested):

simonv3 (approved):

PR #2112, opened 2026-05-17
Minor UI fixes

LouisLouiseKay:
This is all the minor fixes that I didn't know where else to put and didn't want to open a PR for each and every one of them. Some are more visible than others but it's mostly all tweaks and slight improvements.

### DIFF
--- a/client/src/components/common/AddMoneyValueButtons.tsx
+++ b/client/src/components/common/AddMoneyValueButtons.tsx
@@ -1,20 +1,14 @@
 import React from "react";
+
 import Button from "./Button";
 import Money from "./Money";
-import { css } from "@emotion/css";
 
 const AddMoneyValueButtons: React.FC<{
   addMoneyAmount: (val: number) => void;
   currency: string;
 }> = ({ addMoneyAmount, currency }) => {
   return (
-    <div
-      className={css`
-        display: grid;
-        grid-template-columns: repeat(4, 1fr);
-        gap: 0.5rem;
-      `}
-    >
+    <div className="grid grid-cols-4 gap-2 max-md:[&_button]:text-xs! max-md:[&_button]:px-1.5!">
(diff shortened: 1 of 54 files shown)


client/src/components/common/Modal.tsx
```diff
@@ -265,11 +259,11 @@ export const Modal: React.FC<{
 
               <Button
                 className={close}
-                size="compact"
+                onlyIcon
```
timglorioso, line 262: This API feels redundant since Button knows when I'm only providing an icon. Maybe we could test existence of startIcon and endIcon instead?

client/src/components/common/AddMoneyValueButtons.tsx
```diff
@@ -1,20 +1,14 @@
...
 }> = ({ addMoneyAmount, currency }) => {
   return (
-    <div
-      className={css`
-        display: grid;
-        grid-template-columns: repeat(4, 1fr);
-        gap: 0.5rem;
-      `}
-    >
+    <div className="grid grid-cols-4 gap-2 max-md:[&_button]:text-xs! max-md:[&_button]:px-1.5!">
```
timglorioso, line 11: I always worry about smaller buttons on small screens due to touch target (love the the custom property you added earlier! I'd bump it to 50 even but besides the point). Might flex-wrap work here with the original button size?

LouisLouiseKay: As is they just overflowed so that's what I fixed, I can make them wrap instead in a follow up if you prefer !

timglorioso: I see, good to be fixed, no worries

client/src/components/common/Modal.tsx
LouisLouiseKay: Yeah, I guess we could hadn't given it much thought tbh ! This feels a bit out of scope here for all my tiny tweaks? But I'll look into it when I have time.

PR #2115, opened 2026-05-18
Artist can opt in for Federated streaming

MaciAC:
Hello! 

This is a first iteration of the integration between Mirlo and Fairplayer using Canimus as the content data model.

The PR includes:
- A toggle in the artist profile under a feature flag for enabling federated streaming.
- A new endpoint to serve all the artists that opt-in for the federated streaming feature. Allows to filter by "fromDate" which filters newly created and updated artists. The response follows the Canimus data model

I've been using the name "federated streaming" for this interaction among platforms but maybe you all were thinking on a different naming.


src/utils/serialize/artist.ts
```diff
@@ -114,3 +116,75 @@ export const processSingleArtist = (
...
+
+export const serializeSingleArtistIntoCanimus = (artist: LocalArtist) => {
+  const artistUrl = `https://mirlo.space/${artist.urlSlug}`;
+  return {
+    type: "artist",
+    name: artist.name,
+    url: artistUrl,
+    images: {
+      main: {
+        src: `https://cdn.mirlo.space/file/artist-avatars/${artist.avatar?.url[0]}.webp`,
```
MaciAC: is there a standarized better way to define this instead of hardcoding the cdn url?

prisma/migrations/20260501172949_add_federated_streaming_boolean_for_artist/migration.sql
MaciAC, line 1: Not sure if the date of the migration being previous to other already applied migrations will generate any problem on deploy, let me know

simonv3: I don't think so, it's just a way of making sure it's a unique name. the migrations are tracked in a table and applied if not found.

client/src/translation/en.json
```diff
@@ -1201,9 +1201,11 @@
     "deleteImage": "Delete image",
     "warningFeature": "Warning: this is an experimental feature. You shouldn't rely on it.",
     "enableActivityPub": "Enable The Social Web",
+    "enableFederatedStreaming": "Enable federated streaming",
```
simonv3, line 1299: this is a big question but we've turned a bit more towards using stuff like "social web" and "social music" because "federated" is inherently a large and confusing term and we've found that explaining it to people is hard. Taking a queue from Ghost https://ghost.org/help/social-web/

But also we haven't settled on any terminology and are constantly changing wording to try and make it clearer.

src/routers/v1/federatedStreaming/index.ts
```diff
@@ -0,0 +1,103 @@
...
+          avatar: {
+            where: {
+              deletedAt: null,
+            },
+          },
+        },
+      });
+      res.json({
+        type: "root",
+        url: "https://mirlo.space",
```
simonv3: this should probably come from the `process.env.API_DOMAIN` or something similar.

src/utils/serialize/artist.ts
simonv3: You'll wanna construct this `src/utils/images.ts:generateFullStaticImageUrl`

src/utils/serialize/artist.ts
```diff
@@ -114,3 +116,75 @@ export const processSingleArtist = (
...
+  return {
+    type: "album",
+    name: trackGroup.title,
+    url: releaseUrl,
+    release_date: trackGroup.releaseDate,
+    license: trackGroup.credits,
+    artist: trackGroup.artistId,
+    images: {
+      cover: {
+        src: `https://cdn.mirlo.space/file/trackgroup-covers/${trackGroup.cover?.url[0]}.webp`,
```
simonv3: Same as above

src/utils/serialize/artist.ts
```diff
@@ -114,3 +116,75 @@ export const processSingleArtist = (
...
+  const trackUrl = `${releaseUrl}/tracks/${track.id}`;
+  const metadata: any = track.metadata;
+  return {
+    type: "track",
+    name: track.title,
+    url: trackUrl,
+    duration: metadata.format.duration,
+    media: [
+      {
+        src: `https://mirlo.space/v1/tracks/${track.id}/stream/playlist.m3u8`,
```
simonv3: We'll probably want a similar function as above to generate this? And then we could swap it out with the endpoint in the work @txels is doing?

simonv3 (commented):
Looks great!!! Wondering if serializeSingleTrack and serializeSingleTrackGroup should live in respective `/serialize/track.ts` and `/serialize/trackGroup` folders.

src/utils/serialize/artist.ts
txels: https://github.com/funmusicplace/mirlo/pull/2080/changes#diff-eafe7a06db084f84982f44e3557c2a6be216524bc58856d256c4bb792cdfd70b @MaciAC  we're going for a different path where we can handle this flow of "authorized plays".

txels: `https://mirlo.space/v1/tracks/${track.id}/stream/authzd/playlist.m3u8` although open to changing the final URL if another one makes more sense.

client/src/translation/en.json
txels: Haha yes, but you could argue the same about "social" though.

src/routers/v1/federatedStreaming/index.ts
txels: Yes, this will vary per instance.

src/routers/v1/sm/canimus.json/index.ts
```diff
@@ -0,0 +1,103 @@
...
+      if (fromDateFilter) {
+        where.OR = [
+          {
+            createdAt: fromDateFilter,
+          },
+          {
+            updatedAt: fromDateFilter,
+          },
+        ];
+      }
```
txels, line 100: Reminder: figure out how to include also items that have been deleted since the date in the output. It looks like Mirlo uses a consistent `deletedAt` field for that purpose.

client/src/translation/en.json
timglorioso: We can continue to think about naming after this is merged IMO. It's behind a feature flag.

Other stuff I'm thinking about that's relevant to naming: it's clear what streaming toggle does, but the social web toggle now feels too broad. As a user, do I need to have the first for the second to work? Maybe there's a main toggle for social/federation, then separate toggles for posts/etc and streaming? Or maybe all social/federation features are bundled in one toggle? Not sure how much this has been discussed elsewhere, so forgive me if it's noise!

MaciAC: Great point @timglorioso !! I really like the idea of simplifying this interaction into a single toggle that then can be managed more granularly but at first it enables both things :+1:

src/routers/v1/sm/canimus.json/index.ts
MaciAC: This is interesting, 

I am thinking on replacing the boolean stored in db by two datetime columns, someting like: `federatedStreamingOptInDate` and  `federatedStreamingOptOutDate` to manage this part considering not only deleted artists but also the ones that choose to opt out.

MaciAC:
All commens have been applied and this morning we've properly integrated this branch with a local Fairplayer, this can be merged and we will check if the production instances integrate as well

src/routes.ts
```diff
@@ -169,6 +169,7 @@ const routes = [
   "admin/stats",
   "oembed",
   "flag",
+  "federatedStreaming"
```
simonv3: I think this end point is "okay" but I wonder if there's something a bit more precise. In other places we've called endpoints by what it is that they do, so our ActivityPub endpoints are `/ap/:x` routes. I wonder if calling this `canimus` is the most specific and also the most future safe in case we add or stop using it? I don't know and am open to y'alls thoughts @MaciAC @txels @timglorioso @LouisLouiseKay

simonv3 (dismissed):
This looks good to me but I have one question below which is just a 🤔 and I probably could have surfaced earlier. It's not blocking on merging this.

src/routes.ts
txels: Might as well be `canimus.json`

txels: Or `/sm/canimus.json` sm for socialmusic

timglorioso: canimus makes sense to me

MaciAC: absolutely, `/sm/canimus.json` for consistence with other implementations (sockpupet's web) makes sense ! I'll edit it

MaciAC:
I've chaged the endpoint to `sm/canimus.json` as proposed, all references in code are still named "federatedStreaming" as well the toggle in the customize artist page

simonv3 (approved):

PR #2120, opened 2026-05-18
chore: break up cors and apiKey

simonv3:
This splits up the cors and API Key checks into two different middleware, one for the entire app and one for the api.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,19 +1,21 @@
 import express from "express";
 import { initialize } from "express-openapi";
+import slowDown from "express-slow-down";
+import qs from "qs";
 
+import { apiKeyCheck } from "./auth/apiKey";
+import logger from "./logger";
+import rateLimiter from "./rateLimiter";
 import apiDoc from "./routers/v1/api-doc";
 import "./auth/passport";
 
+import routes from "./routes";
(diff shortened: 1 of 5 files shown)


src/auth/apiKey.ts
```diff
@@ -0,0 +1,96 @@
...
+
+import { isValidActivityPubEndpoint } from "../activityPub/utils";
+import logger from "../logger";
+import { AppError } from "../utils/error";
+
+const MIRLO_API_KEY_HEADER = "mirlo-api-key";
+
+const SAFE_METHODS = ["GET", "HEAD", "OPTIONS"];
+
+const isExcludedFromKeyCheck = (path: string, query?: { format?: string }) => {
```
txels, line 12: This is a better name 👍

src/api.ts
```diff
@@ -33,7 +35,7 @@ apiApp.use((req, res, next) => {
   next();
 });
 
-apiApp.use(corsCheck);
+apiApp.use(apiKeyCheck);
 apiApp.use(cookieParser());
```
txels, line 39: BTW I don't think we need the cookieParser in both api and index's middleware stack?

txels (approved):
Thanks for this

PR #2121, opened 2026-05-18
chore: bump timeout for mocha

### LINKED ISSUE #2094: API unit test timeouts
Example from recent deploy: https://github.com/funmusicplace/mirlo/actions/runs/25939108648/job/76251972717

Not sure if failed test #6 is legit, but that part of the code wasn't touched in the deployed changes. Possible separate issue.

simonv3:
close #2094

I think the clearTable on like 40+ tables probably just took too long depending on how resourced the action was.

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -7,7 +7,7 @@ module.exports = {
   spec: "test/**/*.spec.ts",
   require: "ts-node/register",
   forbidOnly: inCI,
-  // watch: !inCI,
+  timeout: 3000,
   "watch-files": ["test/**/*.ts", "src/**/*.ts"],
   "watch-ignore": ["node_modules"],
 };
--- a/test/routers/artists/{id}.spec.ts
+++ b/test/routers/artists/{id}.spec.ts
@@ -23,14 +23,6 @@ describe("artists", () => {
   });
 
   describe("/{id}", () => {
-    beforeEach(async () => {
-      try {
-        await clearTables();
-      } catch (e) {
-        console.error(e);
-      }
-    });
-
     it("should GET /{id} with artist slug", async () => {
       const artistSlug = "test-artist";
       const user = await prisma.user.create({


test/routers/artists/{id}.spec.ts
```diff
@@ -23,14 +23,6 @@ describe("artists", () => {
   });
 
   describe("/{id}", () => {
-    beforeEach(async () => {
-      try {
-        await clearTables();
```
timglorioso, line 28: Thanks for tagging me on this, at least so I can ask questions and learn a bit more context.
1. Why did you only remove from `artists/{id}`? I see several tests timing out: https://github.com/funmusicplace/mirlo/actions/runs/25939108648/job/76251972717
2. It seems like a good idea to `clearTables` on each unit test, but I'm not too versed in testing. Would there be stuff left over from the last test?

(obviously feel free to merge if you know this is an improvement, I'm out of my depth)

.mocharc.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   spec: "test/**/*.spec.ts",
   require: "ts-node/register",
   forbidOnly: inCI,
-  // watch: !inCI,
+  timeout: 3000,
```
timglorioso, line 10: This one makes sense to me! 😄

test/routers/artists/{id}.spec.ts
simonv3: In this specific case clearTables was being called twice for each test, it's called in the `beforeEach` of the parent `describe` too: https://github.com/funmusicplace/mirlo/pull/2121/changes/BASE..d12fadc05d9eb7cade509fd10d3b00af39b37873#diff-6d37034e767e973df13d420476be44c42924ada6cfa2fceef3e5099e3ebcd67dR19

.mocharc.js
simonv3: I'm not sure it'll be enough tbh, and I'm not sure if _the_ fix, but it makes sense to me to try this?

test/routers/artists/{id}.spec.ts
simonv3: (screen shot for easier view 

Image)

PR #2122, opened 2026-05-18
Sales fulfillment sections overhaul

LouisLouiseKay:
This tries to create some more logic around how sections are setup on Mirlo, regrouping Label / Sales / Fullfilment and Account, because these 4 things all work collectively (they even had buttons linking them together because of it), so now you have one space where you can manage artists/label artists, setup you account (to label or not), see your sales and check your order fullfillments.

The sales / fullfilment sections were especially barebone and required some very basic redesign around how tables are handled all over Mirlo (also helps within the admin dashboard).


client/src/components/Account/AccountNav.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { useAuthContext } from "state/AuthContext";
+
+import { bp } from "../../constants";
+
+const AccountNav: React.FC = () => {
+  const { t } = useTranslation("translation");
+  const { user } = useAuthContext();
+  return (
+    <Tabs
+      className={css`
```
simonv3: Tailwind for the future.

client/src/components/common/DateRangeFilter.tsx
```diff
@@ -0,0 +1,123 @@
...
+  | "previousMonth"
+  | "thisYear"
+  | "lastYear";
+
+interface DateRangeFilterProps {
+  value: DateRangeValue;
+  onChange: (value: DateRangeValue) => void;
+}
+
+const DateRangeFilter: React.FC<DateRangeFilterProps> = ({
```
simonv3, line 17: Ah we should re-use this in the admin page too probably.

client/src/components/Sales/SalesRow.tsx
```diff
@@ -78,92 +156,157 @@ const SalesRow: React.FC<{
...
+              }
+            />
+          </Modal>
         )}
       </td>
     </tr>
   );
 };
 
+export const SalesCard: React.FC<{ sale: Sale }> = ({ sale }) => {
```
simonv3: I think this should be its own component (same as FulfillmentCard)

simonv3 (dismissed):
Just think the two Card components should be in their won file.

client/src/components/common/DateRangeFilter.tsx
LouisLouiseKay: We'll have a proper pass on admin, there's a lot of clean up that we can do and lots of potential easy wins I think.

simonv3 (approved):

PR #2124, opened 2026-05-19
chore: temporary files location for local development

txels:
File uploads were not working for me when running Mirlo outside docker due to the location of temporary files.
This uses a more standardised temp file location that makes it possible to run Mirlo both as a container or directly on your machine's OS.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,7 @@ STATIC_MEDIA_HOST=http://localhost:3000
 
 # NGINX_PORT=3001
 MEDIA_LOCATION=./data/media/
+TEMP_LOCATION=/tmp/data/media/
 
 DATABASE_URL="postgresql://nomads:nomads@pgsql:5432/nomads?schema=public"
 POSTGRES_LOCAL_MACHINE_PORT=5434
--- a/src/jobs/generate-album.ts
+++ b/src/jobs/generate-album.ts
@@ -1,8 +1,19 @@
-import { Job } from "bullmq";
-
 import { createReadStream, promises as fsPromises } from "fs";
+import { PassThrough } from "stream";
 
-import { logger } from "./queue-worker";
+import prisma from "@mirlo/prisma";
+import {
+  Artist,
+  Track,
+  TrackArtist,
+  TrackAudio,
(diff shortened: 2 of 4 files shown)


src/jobs/generate-album.ts
```diff
@@ -1,8 +1,19 @@
...
+  Artist,
+  Track,
+  TrackArtist,
+  TrackAudio,
+  TrackGroup,
+  TrackGroupCover,
+} from "@mirlo/prisma/client";
+import archiver from "archiver";
+import { Job } from "bullmq";
+import filenamify from "filenamify";
```
txels, line 15: My IDE sorts typescript imports in a specific way, doesn't look like the repo has specific formatting rules. Sorry if this is annoying/conflict prone... how do you in general deal with these?

simonv3: We added these sorting rules very recently so it's a problem on all PRs, we're just braving the noise on the imports until we've touched every file 😬 . Probably could run a big fix in one go.

.env.example
```diff
@@ -7,6 +7,7 @@ STATIC_MEDIA_HOST=http://localhost:3000
 
 # NGINX_PORT=3001
 MEDIA_LOCATION=./data/media/
+TEMP_LOCATION=/tmp/data/media/
```
simonv3, line 10: Could we default to `/data/media` so that at least it doesn't conflict existing dev environments, or does it have to be `/tmp/`?

txels: It shouldn't conflict I think, `/tmp` should be safe to write to both inside the container and natively on Mac/Linux. These should be short-lived files, there is no mapping of /data to the host (at least in docker-compose-api.yaml). Where do you think this can break something?

txels: `/data/media` OTOH will almost always break when running natively outside a container due to a lack of permissions

src/jobs/generate-album.ts
txels: Yeah tends to be a good idea to bite the bullet once and reformat all.

simonv3 (approved):

PR #2128, opened 2026-05-19
feat: add new route for 'remote' streaming (fairplayer/socialmusic)

txels:
Redoes #2080 after the split of CORS/API done in #2120.


src/routes.ts
```diff
@@ -34,6 +34,7 @@ const routes = [
   "tracks/{id}/trackPlay",
   "tracks/{id}/favorite",
   "tracks/{id}/stream/{segment}",
+  "tracks/{id}/stream/remote/{segment}",
```
txels: @simonv3 at some point you did suggest the `remote` path, after some thought it feels more appropriate than what I used in my prior PR.

txels: ...although we can still change it to whatever makes sense before we ship it.

txels: ` "tracks/{id}/remotestream/{segment}"`
` "tracks/{id}/federatedstream/{segment}"`
...

txels: I have too many suggestions and no strong opinions 😅

simonv3: Maybe `stream/external/{segment}`

Mainly cause "remote" would still include the mirlo app I guess (for which the existence might not make sense anymore if y'all roll out this?) whereas "external" really shows that it's external to the "mirlo" bubble?

src/auth/apiKey.ts
```diff
@@ -41,14 +41,25 @@ export const apiKeyCheck = async (
...
+    const isActivityPub = isValidActivityPubEndpoint(req.path);
+    const urlDoesNotRequireKey = isExcludedFromKeyCheck(req.path, req.query);
+
+    log.debug("API Key check, will pass if any of these is true", {
+      doesNotMutate: !isMutatingRequest,
+      isSameSite,
+      isCORSPreflight,
+      isActivityPub,
+      urlDoesNotRequireKey,
+    });
```
txels, line 53: Once we have validated that the integration works in prod, we may want to remove extra logging. Although probably you don't collect debug logging in prod anyway...?

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
```diff
@@ -0,0 +1,158 @@
...
+            log.debug("Credentials match");
+          }
+        }
+
+        await fetchFile(res, track.audio.id, segment);
+      }
+    } catch (e) {
+      console.error(e);
+      res.status(500);
+      return next();
```
simonv3: Maybe we add some specific cases here? Like, if the `jwt.verify` fails it would end here and it's not really a server error but a verification error right? 

And similarly I'd just `return next(e)` and not set a status, let our generic error handler handle it.

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
```diff
@@ -0,0 +1,158 @@
...
+            });
+            res.setHeader(socialMusic.HEADER_PLAYTOKEN, playToken);
+          }
+        } else {
+          // On a segment request, check that the token matches user and song
+
+          const playToken = req.query[socialMusic.PLAYTOKEN];
+          const reqUserId = req.query[socialMusic.USERID];
+          if (!(playToken && reqUserId)) {
+            res
```
simonv3: This could be replaced with 

```
throw new AppError({
  httpCode: 401,
  description: `${socialMusic.USERID} etc
})
```

And it would go through our normal error handler.

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
```diff
@@ -0,0 +1,158 @@
...
+      }
+
+      const isManifest = segment.includes("playlist.m3u8");
+      if (isManifest && isUserAbleToListenToTrack === "exceeded") {
+        res.status(402).send("Track play limit exceeded");
+        return next();
+      }
+
+      if (track.audio) {
+        const userid = req.get(socialMusic.HEADER_USERID);
```
simonv3, line 66: It looks like `userid` isn't cross referenced with our user id, so I'm wondering if it's a fairplayer userid? I might be misreading it though. Either way worth a comment?

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
```diff
@@ -0,0 +1,158 @@
...
+            return next();
+          }
+
+          if (!userid) {
+            res.status(400).send(`Header ${socialMusic.HEADER_USERID} missing`);
+            return next();
+          }
+
+          const payload = { song: id, userid };
+          if (userid) {
```
simonv3: don't need the `if` here because we `return` in the `if (!userid)` above.

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
```diff
@@ -0,0 +1,158 @@
...
+            log.debug("Generate playtoken:", payload);
+            const playToken = jwt.sign(payload, jwt_secret, {
+              expiresIn: "4w",
```
simonv3: This feels long but i dunno

simonv3 (commented):
Okay, some final thoughts. Style-wise I think maybe throwing the errors through AppError would be more consistent and maybe comment on the userid thing. Other than that it's nitpicks

src/auth/apiKey.ts
simonv3: Yeah the default is set to "info"

And like that here's a way to set it to debug haha https://github.com/funmusicplace/mirlo/pull/2138

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
txels: This makes a lot of sense, a much cleaner pattern.

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
txels: Yes this is a Fairplayer/"external federated entity" userid from the social music protocol, there will be no way to relate those to Mirlo users for the time being until we eventually build some sort of federated identity for the social music (super interesting topic but much longer term IMO).

Adding comment to explain this makes sense. I could even link to the interop docs at the top of the module (I like doing that in my codebases, linking to external docs).

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
txels: Yes, I was also thinking of making this valid for only a few hours.

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
txels: Yeah well spotted, I'll add a test case for that.

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
txels: I think as we go forward standardizing this it would make sense that the userid has a client namespace, e.g. to start with we could provide something like `fairplayer.band:a32698a40cae8a9bc45647eeab54967b7646ce71`.

src/routers/v1/tracks/{id}/stream/external/{segment}.ts
```diff
@@ -0,0 +1,173 @@
...
+              httpCode: 401,
+              description: `${socialMusic.USERID} or ${socialMusic.PLAYTOKEN} missing`,
+            });
+          }
+
+          let decoded;
+          try {
+            // @ts-ignore playtoken type may not be a string?
+            decoded = jwt.verify(playToken, jwt_secret);
+          } catch (e) {
```
txels: @simonv3 I've gone for this rather than discriminate the error type in the catch clause... it seems clearer to me to handle this close to where it happens.

simonv3 (approved):

PR #2130, opened 2026-05-19
User section rework

LouisLouiseKay:
This is a follow up to #2122 to match the logic of two distinct "navmenus". The first includes everything related to "making payments as a consumer" (following/subscribing to artists, purchasing, wishlisting, checking your timeline). The second includes everything related to "receiving payments as an artist" (managing artists, sales, fullfilment, and account settings). This PR focuses on the consumer affordance, mostly grabbing pre-existing systems (two columns layout, pill filtering) and applying them to make the current experience as a listener more unified and coherent.


src/routers/v1/users/{userId}/charges.ts
```diff
@@ -30,15 +32,31 @@ export default function () {
...
+                    },
+                  },
                 },
               },
             },
           },
         },
       });
+      const results = charges.map((c) => ({
+        ...c,
```
simonv3, line 47: For the future probably worth making a charge serializer

client/src/components/Profile/YourPurchases.tsx
```diff
@@ -46,32 +41,20 @@ function YourPurchases() {
...
-            `}
-          >
-            {!purchases || (purchases?.length === 0 && <>{t("noPurchases")}</>)}
-            {purchases && (
+    <div className="p-(--mi-side-paddings-xsmall)">
+      <WidthContainer variant="big" justify="center">
+        <div className="flex flex-wrap md:flex-nowrap gap-8 items-start">
+          <section className="flex-1 w-full min-w-0">
+            <h1>{t("yourPurchases")}</h1>
+            {(!purchases || purchases.length === 0) && (
```
simonv3, line 49: while purchases are loading could probably have a loading screen indicator here. your call though, can be fixed in the future

simonv3 (approved):
lgtm just two optional thoughts

client/src/components/Profile/YourPurchases.tsx
LouisLouiseKay: Ah yeah, I'll add a note for minor fixes, it's something I don't instinctively have in mind to add but you're right (and that page does load quite slowly that it could use it).

PR #2140, opened 2026-05-21
feat(api): load Stripe API key from admin settings with env fallback

### LINKED ISSUE #1147: Store stripe key in database
* [ ] Create a Stripe key value in the settings table
* [ ] Create a UI for this in the admin section
* [ ] Reference this key when doing Stripe things

peterklingelhofer:
Closes #1147


simonv3:
Might take me a minute to review, I'll wanna pull this down and test it.

simonv3:
@peterklingelhofer just FYI  I added so that the API doesn't reveal the key if it's set.

PR #2141, opened 2026-05-21
chore: move to a simplified bucket system for those of us not already…

### LINKED ISSUE #1143: Store bucket names in settings
Right now bucket names are stored as constants in the minio.ts file, but they should be stored in settings. The main reason for this is because if someone were to use backblaze to store files on, they'd have to come up with different bucket names, because backblaze buckets are universal.

simonv3:
… stuck in multi bucket system

Note that this is kind of a one way street (which is desired, we won't want anyone migrating _back_ to legacy after they've enabled the consolidated buckets). 

Note also that the intent is still to switch eventually to the Image table for any of the other things so that we don't have unique Image systems for every type of Image but rather something consolidated. 

Worth doing a little bit more thinking about the impact this will have on CDN caches for images. Might be worth flushing edge caches on cloudflare when this rolls out. 

* [ ] Think about how to handle `clean-up-old-files` job. 

Closes #1143

### DIFF
--- a/src/routers/v1/manage/artists/{artistId}/images/index.ts
+++ b/src/routers/v1/manage/artists/{artistId}/images/index.ts
@@ -34,6 +34,9 @@ export default function () {
     try {
       const { jobId, imageId } = await uploadAndSendToImageQueue(
         { req, res },
+        // imageType "image" = the generic type stored at the root of the
+        // images bucket; model "image" = update the Image table after
+        // optimization (other callers pass e.g. "trackGroupCover").
         "image",
         "image",
         "inFormData",
--- a/src/routers/v1/manage/artists/{artistId}/images/index.ts
+++ b/src/routers/v1/manage/artists/{artistId}/images/index.ts
@@ -1,19 +1,16 @@
+import prisma from "@mirlo/prisma";
+import busboy from "connect-busboy";
 import { NextFunction, Request, Response } from "express";
+
+import { assertLoggedIn } from "../../../../../../auth/getLoggedInUser";
 import {
   artistBelongsToLoggedInUser,
   userAuthenticated,
 } from "../../../../../../auth/passport";
-import { assertLoggedIn } from "../../../../../../auth/getLoggedInUser";
+import { logger } from "../../../../../../logger";
 import { uploadAndSendToImageQueue } from "../../../../../../queues/processImages";
-import busboy from "connect-busboy";
-import prisma from "@mirlo/prisma";
 import { deleteArtistAvatar } from "../../../../../../utils/artist";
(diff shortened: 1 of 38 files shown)


src/routers/v1/manage/artists/{artistId}/images/index.ts
```diff
@@ -37,7 +34,7 @@ export default function () {
     try {
       const { jobId, imageId } = await uploadAndSendToImageQueue(
         { req, res },
-        incomingImageBucket,
+        "image",
         "image",
```
txels, line 41: This twice "image" seems a bit strange. But I guess it comes from having a generic "image" section within the images bucket, if I interpret this correctly.

It would help if there was some README/doc explaining the bucket structure.

src/utils/minio.ts
```diff
@@ -560,3 +674,241 @@ export const removeObjectsFromBucket = async (
...
+// ─── Domain storage operations ────────────────────────────────────────────────
+// All bucket selection and key construction lives here.
+// Call sites express intent; this layer handles the routing.
+
+// ── Audio ─────────────────────────────────────────────────────────────────────
+
+const audioIncomingKey = (audioId: string) =>
+  isConsolidatedMode() ? `incoming/${audioId}` : audioId;
+
+export const uploadIncomingAudio = async (
```
txels, line 788: I like how this breakdown in small explicit functions makes everything more readable.

src/utils/minio.ts
txels, line 1: It doesn't feel like this file is just `minio` though, isn't it valid for any supported object storage?

txels (commented):
This is a good simplification/consolidation.
Some documentation would be super-helpful to better grok the resulting structure.

txels (dismissed):
This is a good simplification/consolidation.
Some documentation would be super-helpful to better grok the resulting structure.

txels:
I am now testing this locally, I set up the bucket prefixes etc. Uploaded an image cover for an album, the worker did a bunch of conversions but it doesn't show in the UI, and shows as broken in the admin:

image

log from task worker: 
```
info: minio: uploading fileStream|buffer: fairplayer-mirlo-images/trackgroup-covers/9488a3bd-924f-4f12-a7e4-265e706bcf4b-x60.png
```

Image tag src: `http://localhost:3001/images/fairplayer-mirlo-images/trackgroup-covers/9488a3bd-924f-4f12-a7e4-265e706bcf4b-x600.webp?updatedAt=2026-06-02T17:23:54.070Z`

The files have been uploaded indeed to minio, I can see them in the console in the expected bucket...
image

But the api console log says:

```
error: minio doesn't have the file fairplayer-mirlo-images/trackgroup-covers {"service":"api"}
```

PR #2143, opened 2026-05-21
Improve layout of release price settings

### LINKED ISSUE #2017: Improve layout of release price settings
The current layout of the price settings on the Edit release page is a bit hard to follow. It makes it look like there's a lot going on, when it's actually fairly simple

## Current

Image

## Suggestion
- Use a simpler arrangement of the settings so each element is consistent - input on the left, help on the right
- Move the banner about subscription tiers to the top so the "0 or donate/Paid/No payment" radio buttons are next to the thing they affect
- Remove the extra percentage label for the  platform cut because it's redundant (i can see how this might be left over from when it was just a slider)
- When "Prefill a suggested contribution" is disabled, just disable the input box and clear its text rather than hiding the whole section
- When "No payments is selected", move the second sentence of the banner to a label below the radio buttons
    - This is the text that reads "When published, this release will not be available for purchase by the public but will be discoverable."
- Simplify heading to just "Price"

### Mockup:
Image

timglorioso:
Close #2017


simonv3 (dismissed):
lgtm

timglorioso:
Thanks @simonv3, hoping to update the e2e test that failed. Lemme know if you're in a hurry to merge

PR #2145, opened 2026-05-22
fix: missing currency in notification feed

simonv3:


src/utils/serialize/trackGroup.ts
```diff
@@ -52,7 +52,11 @@ export const processSingleTrackGroup = (
           : undefined,
       }
     : undefined,
-  merch: tg.merch?.map(processSingleMerch),
+  merch: tg.merch?.map((m) =>
+    processSingleMerch(m, {
+      fallbackCurrency: tg.artist?.user?.currency ?? undefined,
```
simonv3, line 57: Why do we need this? Aren't we drawing from the artist.user.currency in the src/utils/merch?

LouisLouiseKay: This is not for the merch page itself actually, it's for when the merch is nested in a trackgroup. These don't include "merch.artist.user" cause the parent already has access to it, so this is to let the parent pass the currency it already has, otherwise all the nested include have to duplicate "artist.user.currency". (without the fallback it would calculate "undefined" and default to USD in those nested cases)

PR #2159, opened 2026-05-23
fix: don't display 0 suggested price when nonzero min price

timglorioso:


simonv3 (approved):

simonv3:
Just also confirming that the on a new album where a minPrice is set this will set a suggestedPrice if it doesn't exist yet?

timglorioso:
When you create a new album, both UI fields display 0. Both are null in DB. On first save, both are 0 in DB.

For existing albums, regardless of publish status, if min price greater than 0, suggested price displays 0 and is null in DB. This migration makes suggested price match min price in that case.

Does that clarify @simonv3 ?

PR #2168, opened 2026-05-25
Fixed buttons improvements

LouisLouiseKay:
Because the player buttons were made before we added the capacity to wishlist or buy single tracks, if was, often counter intuitively, only possible to wishlist or buy albums through the player buttons, when if you're listening to a lot of things, tracks might be a more immediate candidate in a lot cases (reaching trackcount limit, preparing some favorite tracks, etc.).

I also made the dropdown more accessible on mobile, centered with bigger icons, up until now it could be pretty finnicky to use on smartphones, this makes it much more usable.

### DIFF
--- a/src/utils/serialize/trackGroup.ts
+++ b/src/utils/serialize/trackGroup.ts
@@ -30,13 +30,16 @@ export const processSingleTrackGroup = (
       downloadableContentId: string;
     }[];
     trackGroupPurchases?: { userId: number }[];
+    _count?: { tracks?: number; userTrackGroupPurchases?: number };
   },
   options?: { loggedInUserId?: number }
 ) => {
+  const { _count, ...rest } = tg;
   const currency =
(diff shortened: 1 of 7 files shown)


src/utils/serialize/trackGroup.ts
```diff
@@ -29,43 +29,48 @@ export const processSingleTrackGroup = (
...
-        process.env.API_DOMAIN +
-        `/v1/downloadableContent/${dc.downloadableContentId}`,
-    },
-  })),
-});
+) => {
+  const { _count, ...rest } = tg;
+  return {
+    ...rest,
+    tracksCount: _count?.tracks ?? tg.tracks?.length,
```
simonv3: totalTracks?

simonv3 (dismissed):
Lgtm, just one thing that I think makes more sense wording wise.

PR #2169, opened 2026-05-25
fix: tie trackgroup cover and font size to viewport

LouisLouiseKay:
This became a priority when I started testing Mirlo on smaller / older FHD laptops (on Windows), where the whole layout would be swallowed by the height of the album cover on trackgroup pages which couldn't even be fully displayed. Mostly this is a rework of something I *thought* I'd already done a couple years ago. I've also been working on a similarly more scalable / responsive system for the artist pages.

This shouldn't change anything on most big screen monitor layouts, or OS/Browsers that render pages without any scaling.


client/src/components/TrackGroup/ArtistByLine.tsx
```diff
@@ -27,7 +29,11 @@ const ArtistByLine: React.FC<{
   return (
     <div
       className={css`
-        font-size: 18px;
+        font-size: clamp(
```
simonv3, line 32: Image

simonv3 (approved):

PR #2175, opened 2026-05-26
Artist page viewport fit

LouisLouiseKay:
Follow up to #2169 same issue same conclusion, making a few things more adjustable to smaller desktop/viewport, and adjusting the layout to free up vertical space.

### DIFF
--- a/client/src/components/ManageArtist/ManageArtistDetails/ManageArtistAnnouncement.tsx
+++ b/client/src/components/ManageArtist/ManageArtistDetails/ManageArtistAnnouncement.tsx
@@ -3,17 +3,20 @@ import Button from "components/common/Button";
 import FormComponent from "components/common/FormComponent";
 import MarkdownContent from "components/common/MarkdownContent";
 import React from "react";
+import { createPortal } from "react-dom";
 import { FormProvider, useForm } from "react-hook-form";
 import { useTranslation } from "react-i18next";
 import useArtistQuery from "utils/useArtistQuery";
 import useManagedArtistQuery from "utils/useManagedArtistQuery";
 
-import { bp } from "../../../constants";
+import { between, bp } from "../../../constants";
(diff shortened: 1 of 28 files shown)


client/src/components/ManageArtist/ManageArtistDetails/ManageArtistAnnouncement.tsx
```diff
@@ -66,11 +92,23 @@ const ManageArtistAnnouncement: React.FC<{
 
   const { t } = useTranslation("translation", { keyPrefix: "artistForm" });
 
+  const [mount, setMount] = React.useState<HTMLElement | null>(() =>
```
simonv3, line 99: What does this do and why is it constructed this way? Can we have a comment?

client/src/utils/useFitTitle.ts
```diff
@@ -0,0 +1,68 @@
+import React from "react";
+
+const MOBILE_QUERY = "(max-width: 768px)";
+
+export function useFitTitle<T extends HTMLElement>({
```
simonv3, line 11: Can you add a comment on what this hook does?

client/src/components/ManageArtist/ManageArtistDetails/ManageArtistAnnouncement.tsx
LouisLouiseKay: Yeah, this is because of how the manage page route is handled differently than public artist pages despite "looking similar" and can't use the full width container (tbh that's a rework for another day...), I'll add a comment.

client/src/utils/useFitTitle.ts
LouisLouiseKay: Yeah I'm adding both comments (this one is for balancing very short / very long artist/label names, I've been trying to find a solution for this for a while)

simonv3 (approved):

PR #2179, opened 2026-05-27
Search list restyling

LouisLouiseKay:
This is a follow up to some other work on the search field, but with a significant overhaul visually. Mostly it comes from a frequent issue when trying to look for tracks in the blogpost modal, where the modal's layout would just not work at all with a long list of songs, and I kept battling with z-index issues and the whole modal layout, so I just decided to take the design out of the modal entirely after all, and have a unified search system that just adjust to the context depending on if it's used in the global search or in a blog post. The initial point when I started this eons ago was to just add the possibility to search an artist in the blog post embed search and not just the exact track name, so this also adds that.


simonv3:
Can you attach a picture or screencapture of how it works

client/src/utils/matchesTokens.ts
```diff
@@ -0,0 +1,9 @@
+export const matchesTokens = (
```
simonv3, line 1: Can you add a comment on what this is for

src/routers/v1/tracks/index.ts
```diff
@@ -33,6 +34,34 @@ export default function () {
         where.title = { contains: title, mode: "insensitive" };
       }
 
+      if (q && typeof q === "string") {
+        const tokens = q.split(/\s+/).filter(Boolean);
+        if (tokens.length > 0) {
+          const existingAnd = where.AND
```
simonv3, line 40: Should this almost identical search be re-used?

client/src/utils/matchesTokens.ts
LouisLouiseKay: Yeah sure, this is to avoid refetching an artist collection or wishlist that's already accessible.

simonv3 (approved):

PR #2181, opened 2026-05-28
filter out deleted tracks

MaciAC:
We have found a small issue in the endpoint where we were returning deleted tracks

### DIFF
--- a/src/routers/v1/sm/canimus.json/index.ts
+++ b/src/routers/v1/sm/canimus.json/index.ts
@@ -123,7 +123,7 @@ export default function () {
                   },
                 },
                 include: {
-                  audio: true,
+                  audio: { select: { duration: true } },
                 },
               },
             },
--- a/src/routers/v1/sm/canimus.json/index.ts
+++ b/src/routers/v1/sm/canimus.json/index.ts
@@ -116,10 +116,14 @@ export default function () {
             include: {
               cover: true,
               tracks: {
-                include: {
+                where: {
+                  deletedAt: null,
                   audio: {
-                    select: { duration: true },
+                    uploadState: "SUCCESS",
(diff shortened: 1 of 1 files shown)


src/routers/v1/sm/canimus.json/index.ts
```diff
@@ -116,11 +116,15 @@ export default function () {
...
-                include: {
+                where: {
+                  deletedAt: null,
                   audio: {
-                    select: { duration: true },
+                    uploadState: "SUCCESS",
                   },
                 },
+                include: {
+                  audio: true,
```
txels: Might still be a good idea to limit the number of fetched fields, which is how I interpreted the `select` clause (https://jottup.com/nodejs/optimize-queries-with-prisma-using-include-select-effectively). Since these may end up being large queries that fetch a lot of data.

MaciAC: done!

txels (approved):

src/routers/v1/sm/canimus.json/index.ts
```diff
@@ -116,11 +116,15 @@ export default function () {
...
               cover: true,
               tracks: {
-                include: {
+                where: {
+                  deletedAt: null,
                   audio: {
-                    select: { duration: true },
+                    uploadState: "SUCCESS",
                   },
                 },
```
timglorioso, line 124: This just made me wonder, are we accounting for tracks that are set to "must own"? That is, not streamable unless you've purchased

simonv3: @timglorioso that was added here https://github.com/funmusicplace/mirlo/pull/2128

PR #2183, opened 2026-05-28
Fix UI adjustments

LouisLouiseKay:
Follow up of #2112 and #2160 hopefully the last one of those. It fixes : 

- padding and sizing inconsistencies on mobile
- weird box shadow on chrome
- tables refactor that had broken some elements that I didn't realize used tables
- some labelling issues with things I'd missed 
- lots of "buttons that don't use artist color on artist page"
- added small warnings for stuff that surprised me when going through flows locally
- reworked layouts for "recommended" "successful purchase" etc. those were more "hidden" features I'd never really paid attention to
- reused components to search releases in various sections so that it makes use of the newer picker/search/selectors because it hadn't been updated since.
- added a selector for widget variants when sharing an album
- tried to fix the "modal doesn't appear over limit"... but actually didn't, it's a bigger issue, out of scope and couldn't find what happened.

And some even marginal nitpicks.


client/src/components/Player/BuyTrackModal.tsx
```diff
@@ -1,41 +1,54 @@
...
+import { useTranslation } from "react-i18next";
 
 import Modal from "../common/Modal";
-import { useTranslation } from "react-i18next";
-import { queryTrackGroup } from "queries";
-import { useQuery } from "@tanstack/react-query";
-import BuyTrackGroup from "components/TrackGroup/BuyTrackGroup";
-import { css } from "@emotion/css";
+
+const compactModal = css`
```
simonv3: Couldn't this be a tailwind class? would be simpler?

client/src/components/common/Modal.tsx
```diff
@@ -117,12 +117,47 @@ const Content = styled.div<ContentProps>`
...
+  &:hover:not(:disabled),
+  &:focus,
+  &:focus-visible {
+    background: transparent !important;
+    background-color: transparent !important;
+    filter: none !important;
+  }
+`;
+
+const closeOnLight = css`
```
timglorioso, line 135: Do you have an example of when Modal has a light overlay?

simonv3 (approved):

PR #2187, opened 2026-05-29
feat: include track number in Canimus feed

txels:
I noticed that the track numbers in Fairplayer didn't match those in Mirlo, and the reason was that on the client side we were relying on sort order and the API output wasn't sorted. But I also realised that Canimus actually supports a track number so I added that as well (on the Fairplayer side we have changed ingestion to use track number if present).

I've also relocated the alt link for Canimus to where I think is a more appropriate location than "openGraph/sharing" stuff @simonv3 if that is ok?

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -16,6 +16,17 @@
     -->
     <link rel="manifest" href="/manifest.json" />
     <!--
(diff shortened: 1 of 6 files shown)


simonv3 (approved):
lgtm!

PR #2192, opened 2026-05-31
Add label to artist pages

### LINKED ISSUE #1432: Prominently display label page from artist page.
Display the label affiliation of an artist more prominently on their page and enable links between the two.

LouisLouiseKay:
Closes https://github.com/funmusicplace/mirlo/issues/1432 

This "label badge" is purely opt-in, allows to select amongst any labels an artist might have to display it on their header. Only case where it's created by default is when an artist is created directly by a label account, but it's still easy to opt out.

Displays a super simple modal with radio button on manage page


simonv3:
@LouisLouiseKay will we have to migrate any existing artists that were created directly by a label account?

client/src/components/common/ArtistHeaderSection.tsx
```diff
@@ -256,7 +271,37 @@ const ArtistHeaderSection: React.FC<{
             `}
           />
         )}
-        <div className="w-full flex flex-row items-center justify-end pb-2 max-md:py-1 max-md:border-t max-md:border-(--mi-button-color)/50">
+        <div
```
simonv3: This feels like it should be its own component.

simonv3 (dismissed):
Looks good to me just one clean up thing

LouisLouiseKay:
> @LouisLouiseKay will we have to migrate any existing artists that were created directly by a label account? 

I decided against it, we leave it "as is" for those since they might not be able to immediately see the change and adjust to it any way they want, so for these it's still opt in.

PR #2194, opened 2026-05-31
Tip button in support page

LouisLouiseKay:
The tip button now covers three scenarios which I think represent most use cases : 

- Artist activates the "tip button", but they just have the tip option and no subscription tier, in which case since it has nothing else to display, the support section just opens up a modal with the tip options (now split from the "tip with an incentive to follow" modal because if we're on the tier page or if the artist doesn't have tier, a simple tip modal makes more sense). It also works the same if someone comes from elsewhere on the /support route, and closing the modal lands them on the artist's main page.
- Artist activates the "tip button", and they have tiers : Tip Button appears top right of the support section.
- Artist has tiers but do not want the tip button to appear : The new "tip button" is purely opt in, if you don't activate it, the button doesn't appear in your support section.


prisma/schema.prisma
```diff
@@ -307,7 +307,8 @@ model Artist {
   linksJson                          Json[]
   location                           String?
   allowPurchaseEntireCatalog         Boolean                              @default(false)
-  purchaseEntireCatalogMinPrice      Int                                  @default(0) // in cents  
+  purchaseEntireCatalogMinPrice      Int                                  @default(0) // in cents
+  showTipOnSupportPage               Boolean                              @default(false)
```
simonv3: Honestly I think this should be stored in `properties`. It's a purely display thing, no need for a database change IMO unless there's a good reason for it? Are we ever going to run a query on filtering artists on this? Is it of top level importance?

simonv3 (changes requested):

client/src/components/Artist/Artist.tsx
```diff
@@ -288,10 +307,21 @@ function Artist() {
       </nav>
 
       <ArtistSection>
-        <Outlet />
+        <Outlet context={{ openTipModal }} />
       </ArtistSection>
+      {isTipOnly && (
+        <TipArtistModal
```
simonv3, line 313: ,So this will show on artist albums & posts pages too? 

Edit: looking at screen shot looks like yes

client/src/components/Artist/Artist.tsx
```diff
@@ -288,10 +307,21 @@ function Artist() {
...
+          artist={artist}
+          open={isTipModalOpen}
+          onClose={closeTipModal}
+        />
+      )}
     </>
   );
 }
 
+export type ArtistOutletContext = {
```
simonv3: This feels like it should be in its own file?

prisma/schema.prisma
LouisLouiseKay: You're right, I'd initially gated the whole tip feature so it made more sense, but then I decided to keep it available all the time in the player footer, and for a visibility toggle it doesn't.

simonv3 (approved):

PR #2196, opened 2026-06-01
Artist and trackgroup page adjustments

LouisLouiseKay:
A bunch of chores, slight UI adjustments and last minute fixes, mostly revolving around artist and trackgroup page (the bulk of it around Buy button), with some added chores for policies and tags pages (which didn't have any width container or padding on mobile).

There's an actual bugfix in there though : What we fixed in #1960 on the trackgroup page "Buy" button, was actually still happening when you clicked pre-order or buy on an album that didn't have a stripe account but where the payements were actually received by a label for them. In a similar vein "add to collection" didn't work if someone hadn't setup their stripe account.


client/src/components/Tags.tsx
```diff
@@ -33,19 +33,34 @@ const Tags = () => {
     >
       <SectionHeader>
         <WidthContainer variant="big">
-          <SpaceBetweenDiv>
+          <div
+            className={css`
```
simonv3, line 37: tailwind

simonv3 (dismissed):
lgtm just one small nitpick on tailwind

PR #2190, opened 2026-05-30
Canimus return deleted filter from date

MaciAC:
Hello!

With this PR we introduce:

### List of deleted entitites
We are considering:
- artist that opted out
- artists deleted that opted in at some point
- tracks and trackGroups deleted and related to an artist that has been federated at some point

### fromDate filter
A query string to filter the response by date, we are considering updates and creates:
- artists updated or created from given date (included)
- artists with any track or trackgroup updated or created from given date (included)

for the deleted entities we use deletedAt to filter the response

### other
- fix date format to fit Canimus recommended `YYYY-MM-DD`
- add `updated_date` to all entities

PR #2197, opened 2026-06-02
feat: add support and purchase links to canimus feed

txels:
Ensure that the canimus feed includes a link to the artist's support page.
Additionally fixes the creation of URLs since `join` as a file path function was modifying all `//` into `/` (even for `http[s]://` which ended up with invalid URLs).

Pending adding some test, waiting to sync up with @MaciAC 's tests for https://github.com/funmusicplace/mirlo/pull/2190

### DIFF
--- a/src/utils/serialize/artist.ts
+++ b/src/utils/serialize/artist.ts
@@ -125,8 +125,21 @@ export const processSingleArtist = (
 };
 
 export const serializeSingleArtistIntoCanimus = (artist: LocalArtist) => {
-  const artistUrl = join(String(process.env.API_DOMAIN), artist.urlSlug);
+  const artistUrl = new URL(artist.urlSlug, String(process.env.API_DOMAIN))
+    .href;
   const avatarString = artist.avatar?.url.find((u) => u.includes("x600"));
+  const artistLinks = artist.linksJson.map((link: any) => ({
+    name: link.linkLabel,
+    href: link.url,
+    type: link.linkType,
+    rel: "me",
+  }));
+  artistLinks.unshift({
+    name: "Support",
+    href: new URL("support", `${artistUrl}/`).href,
+    type: "Support",
+    rel: "support",
+  });
 
   return {
     type: "artist",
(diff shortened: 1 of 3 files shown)


src/utils/serialize/artist.ts
```diff
@@ -126,8 +126,21 @@ export const processSingleArtist = (
...
 export const serializeSingleArtistIntoCanimus = (artist: LocalArtist) => {
-  const artistUrl = join(String(process.env.API_DOMAIN), artist.urlSlug);
+  const artistUrl = new URL(artist.urlSlug, String(process.env.API_DOMAIN))
+    .href;
   const avatarString = artist.avatar?.url.find((u) => u.includes("x600"));
+  const artistLinks = artist.linksJson.map((link: any) => ({
+    name: link.linkLabel,
+    href: link.url,
+    type: link.linkType,
+    rel: "me",
```
txels, line 139: Not sure what the links in the Mirlo DB represent at the moment, in terms of relationship, but rel="me" is how it looks to me these are rendered in the UI.

src/utils/serialize/artist.ts
```diff
@@ -126,8 +127,26 @@ export const processSingleArtist = (
...
+  }));
+  const artistSupportsPayment = Boolean(
+    artist.paymentToUser?.stripeAccountId ?? artist.user?.stripeAccountId
```
txels, line 142: @simonv3 is this the correct check?

src/utils/serialize/artist.ts
simonv3: It's these in the right hand side of the UI 

Image

I think they can be a combination of "me" and other things? I'm not sure of the scope of what's allowed in "me"

src/utils/serialize/artist.ts
simonv3: Yes!

src/utils/serialize/artist.ts
```diff
@@ -125,8 +129,26 @@ export const processSingleArtist = (
...
+  );
+  if (artistSupportsPayment) {
+    artistLinks.unshift({
```
simonv3, line 145: This makes sense to me.

simonv3 (approved):
lgtm!

src/utils/serialize/artist.ts
txels: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel those are probaby not all "me"... but I don't think these definitions are super-relevant to us.

Canimus does make a couple up TBH https://github.com/PlaidWeb/Canimus/blob/main/format.md#entity-definitions `support` and `purchase` which we use here.

src/utils/serialize/artist.ts
txels: @simonv3 this is now live and does work fine, however when testing it I realised with some artists direct support will only be enabled if they also have set up subscription tiers.

image

(in that case, in the Mirlo UI there won't be a "support tab" so browser users will never get there)

I see two options for this:
- Just change the copy here, maybe add a sentence saying sthg like "You can still support them by buying their releases.".
- Not include the support link in Canimus in that case. I think this is worse because it will obscure a potential path to end up supporting the artist economically.

cc/ @LouisLouiseKay

PR #2205, opened 2026-06-04
Improve accessibility of header items

timglorioso:
- Remove `p` from donate `button`, which only permits phrasing content ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#technical_summary))
- Ensure screenreader label of search button matches visible label to reflect the same experience
- Fix broken search button tooltip, which is useful when the label is not displayed due to small viewport size (touch devices do not apply)
- Add tooltip to log in link
- Translate menu button label
- Add tooltip to menu button (displayed when logged in)

### DIFF
--- a/client/src/components/Header/Header.tsx
+++ b/client/src/components/Header/Header.tsx
@@ -251,6 +251,7 @@ const Header = () => {
               }
               variant="transparent"
               className={menuButtonOverride}
+              title={t("openMenu", { keyPrefix: "headerMenu" })}
             />
           )}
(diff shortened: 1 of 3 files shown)


LouisLouiseKay:
All good catches, I'm planning a little session with Obigre to catch up on all the strings that have been reframed to better fit context in the past two months so that the new french translation matches as well and this sort of work will help a ton thanks @timglorioso

LouisLouiseKay (approved):

PR #2214, opened 2026-06-06
chore: increase playtoken expiry for federated streaming

txels:
Related to https://codeberg.org/fairplayer/fairplayer/issues/409 - pausing a Mirlo song in Fairplayer to continue listening the next day causes errors because the playtoken has expired. As a mitigation we make the token longer, while we figure out a better way to gracefully handle this at the client's end.

(In the original PR @simonv3 and I briefly discussed token durations but I did not consider that edge case https://github.com/funmusicplace/mirlo/pull/2128#discussion_r3277982924 so I went for too short).


simonv3 (approved):
Sounds good!

PR #2246, opened 2026-06-16
chore: add postmark as a supported mail transport

txels:
Adds Postmark as another of the supported email providers. Also some instructions to make it easier to test the setup in local development.

This is part of the overarching work of making Mirlo easier to self-host by different organizations.


src/jobs/send-mail.ts
```diff
@@ -27,8 +28,9 @@ async function createTransport(): Promise<Transporter> {
       });
     }
 
+    // TODO: some condition for locally testing with a given transport?
```
txels: When trying to test real email sending from a local environment, what approach do Mirlo devs usually take?

simonv3: I guess it could just be "if provider === " and not have the production check. If someone decides to set up a provider locally they should be able too. Historically I just uncomment the line and then recomment it. Thoughts?

simonv3 (dismissed):

src/jobs/send-mail.ts
txels: I am thinking of moving this bit down, as is, as the last "fallback" condition.
This way you have all available setups (settings, env var...) and if all fails and you are in development you still have this.

src/jobs/send-mail.ts
```diff
@@ -257,7 +272,11 @@ export const sendMail = async <T>(job: {
       fromEmail,
     };
 
-    if (process.env.NODE_ENV === "production" || process.env.MAILHOG_PORT) {
+    if (
+      process.env.NODE_ENV === "production" ||
+      process.env.MAILHOG_PORT ||
+      process.env.SEND_EMAILS_IN_DEV === "true"
```
txels, line 278: Haha I had to add this as well as I was going crazy emails were not being sent regardless of configuration.

src/jobs/send-mail.ts
```diff
@@ -27,15 +28,6 @@ async function createTransport(): Promise<Transporter> {
       });
     }
 
-    if (process.env.NODE_ENV !== "production") {
```
txels, line 30: Moved to the end as the last fallback.

simonv3 (approved):

PR #2249, opened 2026-06-17
feat: release art alt text

timglorioso:
- Add field for alt text in release form and DB/API
- Use new field value in places where the art is large enough to see
- Fall back to undefined (new field is optional) where the new field value is used but not added by the artist
- Change alt to empty string in places where art is not large enough to see or primarily functioning as a visual aid
- Display character counter and enforce 250 character maximum

image


simonv3:
Amazing! I'm wondering if there's a way to have it be more clearly tied to the image? Maybe it spans the same width? I realize right now it' snot the most efficient space use how much space the image takes up.

client/src/components/common/ReleaseListSelector.tsx
```diff
@@ -116,7 +116,7 @@ const ReleaseListSelector: React.FC<ReleaseListSelectorProps> = ({
                   {release.cover && (
                     <img
                       src={release.cover.sizes?.[120] ?? release.cover.url?.[0]}
-                      alt={release.title}
+                      alt=""
```
simonv3, line 119: Should these be empty string?

src/utils/serialize/trackGroup.ts
```diff
@@ -126,6 +126,7 @@ export const serializeSingleTrackGroupIntoCanimus = (
       cover: coverString
         ? {
             src: generateFullStaticImageUrl(coverString, finalCoversBucket),
+            alt: trackGroup.coverImageAlt ?? null,
```
simonv3: Is it better to have this be `null` then just `Cover image of Album Title`?

client/src/components/common/ReleaseListSelector.tsx
timglorioso: In this context, the task at hand is picking releases for some purpose. Images here are visual aids, and the real meat of the list is the names of the releases. But when you tab through the list, the first thing announced for each item is the image alt. So when the image alt is the title, you hear "title, artist, title, checkbox not checked", the title being duplicated due to the alt text. This would be even worse if the artist-provided alt text were used, because you'd have to listen through the entire description, then artist (which when you're the artist, you already know), then finally the useful bit, the release title. When the image alt is empty string, you hear "artist, title, checkbox not checked".

src/utils/serialize/trackGroup.ts
timglorioso: I don't think it's useful to specify alt text that duplicates other information, such as the title. I ask myself, what's the value of perceiving a description like that? Does it tell me anything? It tells me there's an cover image for the album, whose title I can probably already find nearby, but I question the value of knowing there's an image if it isn't described.

This probably should be undefined, so the coalesce can be removed.

timglorioso:
> Amazing! I'm wondering if there's a way to have it be more clearly tied to the image? Maybe it spans the same width? I realize right now it' snot the most efficient space use how much space the image takes up.

I like the idea of making it the same width as the image, though for different reasons:

1. This initial design isn't responsive. On small screens, typing into a one-line text field is annoying.
2. For readability, lines of text are best limited to 80 characters. The initial design allows 120 or so.

So we'll make it a textarea with enough rows to hold about 150 characters, which I'll determine by testing how many characters fit in a row at the width of the image on a few different screen sizes.

FWIW I think the size of the image is good, I wouldn't make it smaller.

timglorioso:
@simonv3 since the review, I added a character counter and bumped/enforced the max of 250. I tested this with my band's releases and came just shy for both, so it feels like enough without being [too much](https://adrianroselli.com/2024/04/long-alt.html).

So now the field is responsive for smaller screens and offers a better experience across the board by wrapping text at 72 (ish) characters.

With the explicitness of the field label and hint, and proximity to the image, I don't think we need (or can reasonably do) anything else to "have it be more clearly tied to the image". This simple layout balances complexity of implementation with an experience that should work well regardless of whether someone is using assistive technology.

client/src/components/common/ReleaseListSelector.tsx
simonv3: Nice lesson in "best practices aren't universal"

simonv3 (approved):

PR #2254, opened 2026-06-18
feat(client): indicate available lyrics in player

timglorioso:
achieved by putting lyrics icon next to the now playing track title, which links to the track page which contains the lyrics

image

### DIFF
--- a/client/src/components/Widget/TrackTitleContent.tsx
+++ b/client/src/components/Widget/TrackTitleContent.tsx
@@ -5,6 +5,7 @@ import { Link } from "react-router-dom";
 import { getArtistUrl, getReleaseUrl } from "utils/artist";
 
 import { WidgetLink } from "./utils";
+import { MdLyrics } from "react-icons/md";
 
 export const TrackTitleContent: React.FC<{
   track: Track;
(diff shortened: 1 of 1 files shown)


client/src/components/Widget/TrackTitleContent.tsx
```diff
@@ -59,7 +60,15 @@ export const TrackTitleContent: React.FC<{
   ) : null;
 
   const titleNode = titleLinkTo ? (
-    <Link to={titleLinkTo}>{track.title ?? ""}</Link>
+    <Link className="hover:underline! flex items-center gap-1" to={titleLinkTo}>
```
timglorioso: `flex` here causes a regression: the focus ring isn't visible because the link is larger

simonv3:
Cute! Do we display them anywhere?

timglorioso:
@simonv3 lyrics are already displayed on the track page, which the track title links to. So it felt like a nice way to make that link more informative based on whether a track has lyrics. HAN reported this to help reach [SC 1.2.1](https://www.w3.org/WAI/WCAG22/Understanding/audio-only-and-video-only-prerecorded). There's no mention of music there FWIW.

simonv3 (approved):

PR #2256, opened 2026-06-24
Improve accessibility of release/track cards

timglorioso:
This branch brings several accessibility improvements to the experience of browsing release and track cards (f.k.a. ArtistTrackGroup). I kept the features and UI the same except in a few cases where there were contrast or DOM structure issues.

- Go-to overlay "button" announces title, byline, and the go-to label when focused, and is the first thing to receive focus when tabbing into a card.
- Tab sequence follows visual layout.
- All focus rings are visible and unclipped when using the keyboard. This includes the Buy button.
- Readability of overlay actions improved, both by default and on hover.
- Improved responsive design by making the playable cover art with overlay actions adapt what to display based on their size (using container queries). This avoids cases where the viewport is large enough to squeeze cards into small grid items.

image

https://github.com/user-attachments/assets/efd506e5-559d-4922-8b2d-08e71c7849e5


client/src/pages/:artistId/releases/Index.tsx
```diff
@@ -69,13 +74,16 @@ const ArtistAlbums: React.FC = () => {
...
     >
-      {artist.userId === user?.id && (
+      {isArtistUserLoggedInUser && (
         <SectionActionStrip tight>
           <NewAlbumButton artist={artist} />
         </SectionActionStrip>
       )}
-      {!artist.isLabelProfile && <SortableArtistAlbums />}
-      {artist.isLabelProfile && (
+      {isArtistUserLoggedInUser && !artist.isLabelProfile && (
```
timglorioso, line 82: Narrowed this conditional so that listener UX isn't impacted by the Sortable stuff, which was not improved as part of this branch.

client/src/pages/:artistId/releases/Index.tsx
```diff
@@ -51,10 +53,13 @@ const ArtistAlbums: React.FC = () => {
     return null;
   }
 
-  const hasArtistReleases = (artist.trackGroups?.length ?? 0) > 0;
-  const hasLabelReleases = (releases?.results.length ?? 0) > 0;
+  const releases = artist.isLabelProfile
+    ? labelTrackGroups?.results
+    : artist.trackGroups.map((tg) => ({ ...tg, artist }));
```
timglorioso, line 58: Just a clarifying refactor to consolidate label or artist releases into a single variable. Note the spread of `artist` into each trackgroup, which was borrowed from Sortable (I think). Did not dig deeper on why it's needed, but confirmed that things don't work without it.

client/src/components/Artist/ArtistSquare.tsx
```diff
@@ -1,17 +1,10 @@
...
 import { getArtistUrl } from "utils/artist";
 
-import { bp } from "../../constants";
-
 import ArtistFallbackComposite from "./ArtistFallbackComposite";
-import {
-  TrackGroupWrapper,
-  TrackGroupLinks,
-  TrackGroupInfo,
-} from "./ArtistTrackGroup";
```
timglorioso, line 14: These components were deleted, now inlined in the 2 call sites

client/src/components/Artist/ArtistSquare.tsx
```diff
@@ -42,30 +44,32 @@ const ArtistSquare: React.FC<{
...
-            `}
-          ></div>
-        </TrackGroupLinks>
+            className={cx(
+              "sm:hidden",
+              "[&_button]:-mt-1",
+              "max-sm:[&_button]:-mt-[0.1rem] max-sm:[&_button]:opacity-50"
+            )}
+          />
+        </div>
```
timglorioso: Not sure what this empty div is doing, so I kept it.

client/src/components/common/PlayableCover.tsx
```diff
@@ -0,0 +1,393 @@
...
+ * @param trackIds - Playable track IDs queued when the play button is clicked.
+ * @param title - Alt text for the cover image.
+ * @param image - Cover image descriptor from {@link buildCoverImage}.
+ * @param goToId - `id` for the screen-reader-only "go to" label.
+ * @param goToLabel - Visible overlay link text (e.g. "Go to album").
+ * @param goToUrl - Destination for the overlay link.
+ * @param overlayActions - Extra controls rendered in the overlay action bar.
+ * @param className - Applied to the cover root element.
+ */
+function PlayableCover({
```
timglorioso, line 269: A replacement for ClickToPlay with revised API and enhanced accessibility.

client/src/components/common/ReleaseCard.tsx
```diff
@@ -0,0 +1,150 @@
...
+ * {@link PlayableCover}, a title link, optional artist byline, fundraiser
+ * thermometer, and purchase/wishlist overlay actions.
+ *
+ * @param trackGroup - The album or single to display.
+ * @param showArtist - Whether to show the artist name below the title.
+ * @param showWishlist - Whether to show the wishlist button for logged-in users.
+ * @param headingLevel - Semantic heading level for the title link.
+ * @param as - Root element type; defaults to `li` for list layouts.
+ */
+function ReleaseCard({
```
timglorioso, line 30: A replacement for ArtistTrackGroup with revised API and enhanced accessibility.

client/src/components/common/TrackCard.tsx
```diff
@@ -0,0 +1,114 @@
...
+ * {@link PlayableCover}, a title link, optional artist byline, and
+ * purchase/wishlist overlay actions.
+ *
+ * @param track - The track to display (includes its parent `trackGroup`).
+ * @param showArtist - Whether to show the artist name below the title.
+ * @param showTrackWishlist - Whether to show the per-track wishlist button.
+ * @param headingLevel - Semantic heading level for the title link.
+ * @param as - Root element type; defaults to `li` for list layouts.
+ */
+function TrackCard({
```
timglorioso, line 26: I split ArtistTrackGroup into ReleaseCard and TrackCard (this one) to keep the API's clean and to focus on the domain-centered dev problem "I have a release and I want to display it"

simonv3:
I'd like @LouisLouiseKay to have a look at this when they have chance cause they're most familiar with this

client/src/components/common/PlayableCover.tsx
LouisLouiseKay: ClickToPlay was always kind of a bad name for that one so definitely happy to see it replaced.

client/src/components/common/ReleaseCard.tsx
LouisLouiseKay: Had less issues with that one, but it's cleaner too that way anyway.

client/src/components/Artist/ArtistSquare.tsx
LouisLouiseKay: So now we need to edit in both if we make a change? Could this be some shared code to make it easier to edit in the future? It's not a big deal though but yeah.

client/src/components/Artist/ArtistSquare.tsx
LouisLouiseKay: I think it's dead code we can delete it.

client/src/components/common/PlayableCover.tsx
LouisLouiseKay: btw any reason why we don't delete ClickToPlay same as ArtistTrackGroup?

client/src/components/Artist/ArtistSquare.tsx
timglorioso: Yeah. It's a tradeoff. In my experience, some things aren't worth the trouble of making reusable because you end up editing it for one specific context, and unknowingly breaking it in other contexts. Of course, the alternative here is that you might want to edit it in all contexts (would love to get Storybook more fleshed out to visualize them) and not know where the other places are to accomplish that.

There's also diminishing returns on reusing small bits like this that tend to be context-aware anyway. Like, where am I supposed to use TrackGroupInfo? Would I use it on a track group page? In other contexts, the styles and DOM structure often diverge, and in my experience it's cleaner to copy and edit than to make conditionals within the existing component.

I'm also generally trying to avoid importing bits from components in other parts of the file hierarchy because it can get a little spaghetti for my taste.

client/src/components/common/PlayableCover.tsx
timglorioso: Good catch, removed

PR #2264, opened 2026-06-26
Client pages directory

timglorioso:
A client refactor that moved about 90 files from components into a new pages directory. Those 90 files were moved because the routes file imports them. The proposed convention is that anything imported by the routes file goes in pages with a subdirectory structure matching route paths, similar to the API. It's just an organizational strategy, there are no mechanics added to enforce.

I also updated route paths to use kebab-case in traditional web form: 3 admin routes and bulk track upload, so minimal impact to bookmarks etc. and no need for redirects.


client/src/components/ManageArtist/ManageArtistDetails/CustomNamesForTabs.tsx
```diff
@@ -5,7 +5,7 @@ import { useFormContext } from "react-hook-form";
 import { useTranslation } from "react-i18next";
 import useManagedArtistQuery from "utils/useManagedArtistQuery";
 
-import { ArtistFormData, ArtistFormSection } from "./CustomizeLook";
+import { ArtistFormData, ArtistFormSection } from "pages/manage/artists/customize/Index";
```
simonv3: My main concern with this kind of file rename (using `Index`) is that now there's dozens of files named Index and searching for the right one is going to be slightly more difficult. Like a search for the component name itself when it gets imported doesn't show up in results anymore, you have to rely on path results. I was also going to say that it also implies that there is no consistent guaranteed name for the component when it gets called, but I guess that's less important because this is for routes?

client/src/components/ManageArtist/ManageArtistDetails/SortableTabsOrder.tsx
```diff
@@ -17,7 +17,7 @@ import { useTranslation } from "react-i18next";
 import { TabId } from "utils/artistTabs";
 import useManagedArtistQuery from "utils/useManagedArtistQuery";
 
-import type { ArtistFormData } from "./CustomizeLook";
+import type { ArtistFormData } from "pages/manage/artists/customize/Index";
```
simonv3: Just curiuos what does importing it as a `type` gain us here? I hadn't seen this before last week.

client/src/components/FulFillment/Fulfillment.tsx
```diff
@@ -1,157 +0,0 @@
-import { useQuery } from "@tanstack/react-query";
```
simonv3, line 1: Ah it's a shame that git doesn't recognize this as a rename, because then I think we'd easier preserve git history? whereas right now with the new file I don't think it will have a history.

client/src/components/ManageArtist/ManageArtistDetails/CustomNamesForTabs.tsx
timglorioso: Since it's similar to the API naming, I'm curious if you have difficulty searching there?

> Like a search for the component name itself when it gets imported doesn't show up in results anymore, you have to rely on path results.

These components should generally only be imported by routes.tsx.

I explored other options first, like reusing the last segment of the path, but it felt more confusing. Borrowing the classic "index.html" naming felt recognizable and the regularity felt easier to distinguish alongside other components like Layout.

client/src/components/ManageArtist/ManageArtistDetails/SortableTabsOrder.tsx
timglorioso: Looks like it was already there, so not sure why, but reading up on it seems to be useful if we have isolatedModules enabled (we do) and we re-export imported types (we don't?).

client/src/components/FulFillment/Fulfillment.tsx
timglorioso: Fixed

client/src/components/ManageArtist/ManageArtistDetails/CustomNamesForTabs.tsx
simonv3: > Since it's similar to the API naming, I'm curious if you have difficulty searching there?

Not really, and that's because in the API I'm looking for a route. I don't know that I have that clear semantic tie in between a page and a route on the client. Mainly because components are more versatile than just "pages". I'll also admit that I don't really browse by folder structure for components, like, at all, it's all by search of the name. From experience, it feels like Vscode seems to have a harder time surfacing Profile/index to the top of the search for a specific component name.  Maybe the folder structure will help with that? 

Edit: ah I see now you added folder structures that nest deeper to match the routes.

client/src/components/Profile/index.tsx
```diff
@@ -1,82 +1,7 @@
...
-import Pill from "components/common/Pill";
-import TrackgroupGrid from "components/common/TrackgroupGrid";
-import WidthContainer from "components/common/WidthContainer";
-import React from "react";
-import { useTranslation } from "react-i18next";
-import { useAuthContext } from "state/AuthContext";
-
-import FilterGroup from "./UserNotificationFeed/FilterGroup";
 
 export const ProfileSection = styled.div`
```
simonv3: If this file is just ProfileSection it should probably be called that?

simonv3 (dismissed):

client/src/components/Profile/index.tsx
timglorioso: Done

client/src/components/ManageArtist/ManageArtistDetails/CustomNamesForTabs.tsx
timglorioso: Yeah I'm starting to doubt this a little? It felt right to me, and I actually pared it down from a larger refactor with `components/domain` etc. alongside `components/common`. The thing is, routes.tsx already tells us everything that this new pages dir contains. So maybe this part feels less valuable in isolation?

I can't put my finger on it, maybe part of it is having a less overwhelmingly large components directory, another part being able to see all the top-level route components in the file hierarchy without having to open routes.tsx and mental map from path to another file... the component organization overall feels a little loosey-goosey right now and I'm wanting more structure, as someone who didn't write much of it originally.

simonv3: Yeah I think that's fair and I think it probably comes from just a different way of navigating the code, which I do almost entirely with search. I don't think this is a bad idea! Since you've already done the work, feels like it's good to merge (and probably sooner rather than later to avoid more and more conflicts).

PR #2290, opened 2026-07-03
High contrast mode

timglorioso:


client/src/styles/theme.css
timglorioso, line 1: This got kind of horrifying but I also feel like we generally need a big styles refactor, at which point there may as well be some design reconsiderations anyway. Hoping to add least get the class selectors into specific components so there's less magic global styles from unknown places.

PR #2307, opened 2026-07-09
feat: toggle to not send every renewal and just monthly income

simonv3:
Very motivated by my own need for this to manage my inbox lol


timglorioso:
I feel like this would even be a sensible default instead of having a toggle for it. After a couple subscribers, it probably gets noisy enough to want to combine.

simonv3:
Right, I think for now though I want to leave it as is, I think it some people (eg CST) rely on it for accounting purposes.

PR #2309, opened 2026-07-09
Artist-to-Profile 1: clearing current use of "profile"

timglorioso:
refactor: move profile to listener
- (api) GET /auth/profile -> GET /auth/user
- (client) /profile/* -> /listener/* (includes temporary client redirects)
- (client) pages/profile/ -> pages/listener/
- (client) components/Profile/ -> components/Listener/
- (client) translation keys profile -> listener
- (client) Cypress / vitest updated paths and auth/user intercepts


client/src/components/Header/Menu.tsx
```diff
@@ -86,13 +86,13 @@ const Menu = forwardRef<
         <nav aria-label={t("main")} className="flex-auto">
           <ul>
             <li>
-              <MenuLink onClick={onClose} to="/profile/notifications">
```
simonv3, line 89: Are you picturing a different notifications page for a listener than for an artist?

client/src/pages/manage/artists/{artistId}/posts/Index.tsx
```diff
@@ -1,7 +1,7 @@
 import { useQuery } from "@tanstack/react-query";
 import { ArtistButton } from "components/Artist/ArtistButtons";
 import SectionActionStrip from "components/common/SectionActionStrip";
-import FilterGroup from "components/Profile/UserNotificationFeed/FilterGroup";
+import FilterGroup from "components/Listener/UserNotificationFeed/FilterGroup";
```
simonv3, line 4: Maybe not for this PR but maybe these kind of things need a more common folder?

simonv3 (commented):
Ah interesting. This feels a little like it's setting up the Listener section for the listener profile, but there's a lot on Listener that I don't think should track to a Listener profile (like, "plays"). But it makes sense I think. 

I feel like it's not necessarily a blanket switch from everything in Profile -> everything in Listener.

timglorioso:
My goal with this branch is to make the use of "profile" unambiguous. So we could use "user" instead of "listener" as the interim term if it's less confusing with that future listener work.

client/src/components/Header/Menu.tsx
timglorioso: We could keep user-facing profile URLs and live with the ambiguity until listener profiles exist.

client/src/pages/manage/artists/{artistId}/posts/Index.tsx
timglorioso: Trust me, I have plans for more component organization :) Today is not the day

simonv3:
@timglorioso yeah that sounds good to me.

timglorioso:
@simonv3 I'm thinking the best way forward is to keep existing profile stuff as-is (so basically scrapping this commit) and live with the ambiguity of "profile" as User and "profile" as Profile (fka Artist) until we have listener profiles. Because otherwise the `/profile/*` URLs would change twice: once now, and once later when we do listener profiles.

I think regardless /auth/profile should become /auth/user.

I'll close this PR in the meantime.

PR #2312, opened 2026-07-09
feat: add umami as a cookie-less on-site tracker

simonv3:
@timglorioso @LouisLouiseKay what do you think? this adds our own [umami](https://umami.is/) instance for analytics. Things are stored on our own postgres database (shared with the main database). It would bump our monthly cost $7/mo with the starter render instance. Seems worth it to me to get insight of where traffic is coming from. Note that this is different from analytics for artists which I think is something we should build out ourselves rather than using a third party instance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules/
 dist/
+*.tsbuildinfo
 .env
 /data
 error.log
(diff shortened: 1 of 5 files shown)


timglorioso:
Down to give it a try

PR #2313, opened 2026-07-09
refactor: db+api Artist renamed to Profile

timglorioso:
Tables renamed in schema and where referenced by `prisma.*`. No change to attribute names, only attribute types. No change to API endpoints.

### DIFF
--- a/src/routers/v1/manage/purchases/index.ts
+++ b/src/routers/v1/manage/purchases/index.ts
@@ -34,7 +34,7 @@ export default function () {
     try {
       if (!artistIds) {
         artistIds = (
-          await prisma.artist.findMany({
+          await prisma.profile.findMany({
             where: {
               userId: user.id,
             },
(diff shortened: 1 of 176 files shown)


src/routers/v1/manage/purchases/index.ts
```diff
@@ -110,7 +110,7 @@ export default function () {
 
 const csvColumns = [
   {
-    label: "Artist",
```
timglorioso, line 113: Flagging this as I'm unsure if anyone expects this in CSVs

src/routers/v1/manage/sales.ts
```diff
@@ -180,9 +180,9 @@ export default function () {
 const csvColumns = [
   { label: "User-Friendly ID", value: "userFriendlyId" },
   { label: "Date", value: "datePurchased" },
-  { label: "Artist", value: "artist" },
+  { label: "Profile", value: "artist" },
   { label: "Type", value: "saleType" },
-  { label: "Artist", value: "artist" },
```
timglorioso, line 185: Flagging this as I'm unsure if anyone expects this in CSVs

PR #2317, opened 2026-07-11
Artist-to-Profile Internal 1: Tables

timglorioso:
Only "Tables" changes for review. Prisma models (tables) renamed. No change to fields and beyond.

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -49,55 +49,55 @@ model Client {
 }
 
 model User {
-  id                                  Int                             @id @default(autoincrement())
-  email                               String                          @unique
+  id                                  Int                              @id @default(autoincrement())
+  email                               String                           @unique
   name                                String?
-  password                            String                          @default("")
+  password                            String                           @default("")
   refresh                             String?
-  createdAt                           DateTime                        @default(now())
-  updatedAt                           DateTime                        @default(now()) @updatedAt
+  createdAt                           DateTime                         @default(now())
+  updatedAt                           DateTime                         @default(now()) @updatedAt
   deletedAt                           DateTime?
-  isAdmin                             Boolean                         @default(false)
+  isAdmin                             Boolean                          @default(false)
   countryCode                         String?
   currency                            String?
-  emailConfirmationToken              String?                         @default(uuid()) @db.Uuid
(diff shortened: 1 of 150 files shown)


src/jobs/send-out-monthy-income-report.ts
```diff
@@ -112,7 +123,7 @@ const sendOutMonthlyIncomeReport = async () => {
     ]);
     const buyers = keyBy(buyerRows, "id");
     const groupedCancellations = groupBy(
-      cancelledSubscriptions.map((subscription) => ({
+      cancelledSubscriptions.map((subscription: CancelledSubscriptionRow) => ({
```
timglorioso: reconsider

src/routers/v1/admin/artists/index.ts
```diff
@@ -67,7 +67,7 @@ export default function () {
         schema: {
           type: "array",
           items: {
-            $ref: "#/definitions/Artist",
+            $ref: "#/definitions/Profile",
```
timglorioso: This feels not desirable because the external contract is still Artist-based. Would keeping Artist break stuff? It's docs but I saw recently a wrong type here that caused a runtime bug, something like a payment amount string vs number.

Check the api-doc file and confirm that these $ref need to match and don't impact runtime behavior.

simonv3: I think this name isn't meaningful outside of just the docs.

src/jobs/send-out-monthy-income-report.ts
simonv3: Why reconsider?

simonv3 (dismissed):

src/jobs/send-out-monthy-income-report.ts
timglorioso: It was a type I added to try to fix some test failures, but I wasn't confident that it's actually needed.

prisma/schema.prisma
timglorioso, line 1: I ran prisma format and it added a single space on a bunch of lines, so the diff is unfortunately messy

src/jobs/send-out-monthy-income-report.ts
timglorioso: Yeah everything's fine without it

PR #2318, opened 2026-07-11
Artist-to-Profile Internal 2: Fields

timglorioso:
Prisma model fields renamed. API consumers and email templates don't see any changes. API requests map params from artist to profile. API responses and email jobs use updated serializers to map from profile to artist.


prisma/migrations/20260717160340_rename_fields_artist_to_profile_generated/migration.sql
```diff
@@ -0,0 +1,74 @@
+-- RenameForeignKey
```
simonv3: Oh I think I merged this? Or something very similar to it? Is it still needed?

simonv3 (commented):
I wonder if a lot of the maps that are happened should actually be happening in a `serializer` which is after all the formatter of the data. Adding those serializers for controllers that are missing them would be a good move overall. What do you think @timglorioso?

prisma/migrations/20260717160340_rename_fields_artist_to_profile_generated/migration.sql
timglorioso: This handles the field renames e.g. `artistId` to `profileId` in the constraints. Yours caught the constraints from the table renames.

timglorioso:
@simonv3 thanks for checking in. Sounds good. I was figuring on adding integration tests here to verify this PR too.

emails/announce-label-invite/subject.pug
```diff
@@ -1,2 +1,2 @@
-- const labelName = (labelArtist && labelArtist.name) || (label && label.name) || "a label"
```
timglorioso, line 1: Just a tweak because labelArtist and label were both getting passed the same label artist/profile in src/routers/v1/manage/artists/{artistId}/labels/index.ts.

src/routers/v1/flag.ts
```diff
@@ -55,7 +56,9 @@ export default function () {
             reason,
             description,
             trackGroupId,
-            trackGroup,
+            trackGroup: trackGroup
```
simonv3, line 59: We probably shouldn't send an e-mail at all if there's no trackGroup right? I guess out of the scope of this PR.

timglorioso: Yeah the logic in this endpoint is a little funky. I feel like there could be a bit more of a systematic flow across the board. Prefer guard-esque statements so it's linear instead of branching and nesting.

src/routers/v1/artists/{id}/confirmFollow.ts
```diff
@@ -70,7 +70,7 @@ export default function () {
 
         const settings = await getSiteSettings();
         const instanceArtistId = Number(
-          settings.settings?.instanceCustomization?.artistId
+          settings.settings?.instanceCustomization?.profileId
```
simonv3: Did we actually migrate this in the settings? This might take a transform in the blob during the migration.

src/routers/v1/artists/{id}/follow.ts
```diff
@@ -66,7 +66,7 @@ export default function () {
 
       const settings = await getSiteSettings();
       const instanceArtistId = Number(
-        settings.settings?.instanceCustomization?.artistId
+        settings.settings?.instanceCustomization?.profileId
```
simonv3: Same thought about the settings migration.

src/routers/auth/signup.ts
```diff
@@ -144,15 +144,15 @@ const signup = async (req: Request, res: Response, next: NextFunction) => {
       if (receiveMailingList) {
         const settings = await prisma.settings.findFirst();
         if (
-          settings?.settings?.instanceCustomization?.artistId &&
+          settings?.settings?.instanceCustomization?.profileId &&
```
simonv3: Same comment about the settings migration.

src/serializers/notification.ts
```diff
@@ -0,0 +1,179 @@
...
+): Serialized<T> => {
+  const {
+    profileId,
+    profile,
+    subscription,
+    post,
+    trackGroup,
+    relatedUser,
+    relatedUserId,
+    trackGroupId,
```
simonv3: Do these two need to be added back in line 82?

simonv3 (approved):
I think this is good to go but I wish we had a way to test this on a stage migration.

simonv3 (changes requested):
i think the instance profile thing needs to be adjusted

src/serializers/notification.ts
timglorioso: They were only used to match purchases, so I moved that logic back into the `/users/{userId}/notifications` endpoint and removed those variables from the destructure.

src/routers/v1/artists/{id}/confirmFollow.ts
timglorioso: Reverted. I'll handle settings in a later PR. Added tests to catch this moving forward: test/settings-json-keys.spec.ts, test/utils/settings.spec.ts

PR #2330, opened 2026-07-16
fix(client): player muted by default

timglorioso:
Firefox Nightly respects the `muted` attribute on the `video` element, which appears to be set the second time you press play anywhere on the site.

It looks like this was intentional in cda9189366cfee86169e5aa5cb136d2de1df6002. @simonv3 I'm not sure what purpose the `muted` attribute is serving here, do you have any idea?

### DIFF
--- a/client/src/components/Player/AudioWrapper.tsx
+++ b/client/src/components/Player/AudioWrapper.tsx
@@ -152,7 +152,6 @@ export const AudioWrapper: React.FC<{
               playerRef.current.pause();
             }
           } else {
-            playerRef.current.setAttribute("muted", "");
             playerRef.current.playsInline = true;
             playerRef.current.play().catch(() => {
               dispatch({ type: "setPlaying", playing: false });


simonv3:
Yeah so IIRC (and this was 2023, mind you) I think this is related to the fact that because we play our audio element through an interface that isn't directly the audio element (ie, the user is pressing play somewhere else on the website) for mobile browsers this counts as non direct trigger of the play and as a result is an autoplay, which some browsers (IIRC it was safari on mobile) didn't like. It had to start muted. So I think that's what this was trying to fix. It's possible there's a better solution! If this works on Safari iOS / chrome / etc seems worth shipping.

timglorioso:
The element triggering the event can be anywhere in the DOM. The culprit is usually extra function calls between the event handler and `play()`. I actually think browsers have gotten better with detecting when there's a valid origin event, but it would have been 2021 or so the last time I had hands-on experience with those playback errors.

iOS Safari seems okay. I managed to test by getting my dev env serving on my router's local network. Had to build the client for SSR and change the client API env var to my "server"'s IP instead of localhost.

I'll test after it's deployed too.

simonv3 (approved):

PR #2341, opened 2026-07-24
fix: improvements to which data shows in federated feeds

txels:
The Canimus format itself may end up being short lived, as we will soon transition to a new version of the protocol, but these filtering conditions will apply anyway:

- Do not include in the feed artists that are actually labels
- Include the internal immutable Mirlo uid for entities so the receiving end can more easily match indexed and remote entities
- Filter out artists that have no released music
- Filter out songs that are not available for "preview" i.e free listen

### DIFF
--- a/prisma/seeds/artists.ts
+++ b/prisma/seeds/artists.ts
@@ -16,6 +16,8 @@ export const artists: Prisma.ProfileCreateInput[] = [
         linkType: "Website",
       },
     ],
+    federatedStreaming: true,
+    federatedStreamingOptInDate: new Date("2026-07-12T08:00:00Z"),
   },
   {
     name: "Robin",
--- a/src/routers/v1/sm/canimus.json/index.ts
+++ b/src/routers/v1/sm/canimus.json/index.ts
@@ -111,23 +111,47 @@ export default function () {
         createdAt: "desc",
       };
 
+      const validTracks: Prisma.TrackWhereInput = {
+        deletedAt: null,
+        isPreview: true,
+        audio: {
+          uploadState: "SUCCESS",
+        },
+      };
+
+      const validTrackGroups: Prisma.TrackGroupWhereInput =
+        whereForPublishedTrackGroups();
+
       const artists = await prisma.profile.findMany({
-        where: federatedArtist,
+        where: {
+          ...federatedArtist,
+          // only if they have some trackGroup
+          trackGroups: {
+            some: {
+              ...validTrackGroups,
+              // only if they have some track
+              tracks: {
+                some: validTracks,
+              },
+            },
+          },
+        },
         skip: skipQuery ? Number(skipQuery) : undefined,
         take: take ? Number(take) : undefined,
         orderBy: orderByClause as Prisma.ProfileOrderByWithRelationInput,
         include: {
           trackGroups: {
-            where: whereForPublishedTrackGroups(),
+            where: {
+              ...validTrackGroups,
+              // only if they have some track
+              tracks: {
+                some: validTracks,
+              },
+            },
             include: {
               cover: true,
               tracks: {
-                where: {
-                  deletedAt: null,
-                  audio: {
-                    uploadState: "SUCCESS",
-                  },
-                },
+                where: validTracks,
                 include: {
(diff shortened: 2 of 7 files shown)


prisma/seeds/artists.ts
```diff
@@ -16,6 +16,8 @@ export const artists: Prisma.ProfileCreateInput[] = [
         linkType: "Website",
       },
     ],
+    federatedStreaming: true,
+    federatedStreamingOptInDate: new Date("2026-07-12T08:00:00Z"),
```
txels, line 20: I find it convenient to at least have one artist in the seed that is available for federated streaming.

src/routers/v1/sm/canimus.json/index.ts
```diff
@@ -111,23 +111,47 @@ export default function () {
         createdAt: "desc",
       };
 
+      const validTracks: Prisma.TrackWhereInput = {
+        deletedAt: null,
+        isPreview: true,
```
txels: This is up for debate, but I think it makes sense that music that shouldn't be available for "free listens" at all should not be part of the feed.

Alternatively, for future versions of the feed, we may consider including these tracks but not the links to their audio, so they are available for search/discovery but not to listen to.

src/routers/v1/sm/canimus.json/index.ts
```diff
@@ -111,23 +111,47 @@ export default function () {
...
+        where: {
+          ...federatedArtist,
+          // only if they have some trackGroup
+          trackGroups: {
+            some: {
+              ...validTrackGroups,
+              // only if they have some track
+              tracks: {
+                some: validTracks,
+              },
```
txels: I haven't found how to tell prisma to filter out artists where the join condition will return no trackgroups  without duplicating this. I presume we would need a more structured query for that, use subqueries... if anyone knows a quick solution let me know.

src/utils/artist.ts
```diff
@@ -155,14 +155,17 @@ export const whereForAllArtistsThisLabelCanAddReleasesFor = (
 
 export const artistDeleted: Prisma.ProfileWhereInput = {
   deletedAt: { not: null },
+  isLabelProfile: false,
```
simonv3: Labels could have compilations or similar on their profile, so they might have music.

src/routers/v1/sm/canimus.json/index.ts
timglorioso: Might make more sense to handle on the Fairplayer side. If this endpoint gets used by other parties, they might have different desires of handling non-playable catalog.

txels: Hmm so then @timglorioso maybe you'd be OK with the second option, include it in the feed but without a link to the audio file?

src/utils/artist.ts
txels: Ah, good point, I wasn't considering this, it makes a lot of sense.

src/routers/v1/sm/canimus.json/index.ts
txels: I've gone for the "export but without media" option, let me know if you have an objection.

src/utils/artist.ts
txels: Solved.

src/routers/v1/sm/canimus.json/index.ts
simonv3: I don't know of a fix for it prisma-wise but you could copy the object into its own `const` and re-use rather than duplicate?

simonv3 (dismissed):

src/routers/v1/sm/canimus.json/index.ts
txels: Sure, I was hoping for some magical way to cross-reference the clause but a const will do for now.

txels: Done [here](https://github.com/funmusicplace/mirlo/pull/2341/changes/1115c2eee9fa0aacef46f1746249dcd9fb8e24a5)

simonv3 (approved):

PR #2359, opened 2026-08-01
Artist-to-Profile Internal 3: helpers locals

timglorioso:
Aliasing shared helpers so upcoming locals branches can independently switch to the renames. The final locals branch will remove "artist" aliases.


simonv3 (dismissed):

src/routers/auth/profile.ts
```diff
@@ -1,25 +0,0 @@
-import prisma from "@mirlo/prisma";
-import { NextFunction, Request, Response } from "express";
-
-import {
-  USER_PROFILE_SELECT,
-  serializeUserProfile,
-} from "../../serializers/userProfile";
-
-const profile = async (req: Request, res: Response, next: NextFunction) => {
```
timglorioso, line 9: Inlined this handler in auth/index.ts, the only caller

src/serializers/userProfile.ts
```diff
@@ -1,155 +0,0 @@
...
-  finalArtistAvatarBucket,
-  finalCoversBucket,
-  finalUserAvatarBucket,
-  finalUserBannerBucket,
-} from "../utils/minio";
-
-import { omitApPrivateKey, Serialized } from "./utils";
-import { serializeProfileUserSubscription } from "./profileUserSubscription";
-
-export const USER_PROFILE_SELECT = {
```
timglorioso, line 14: Moved/renamed to userSelect in utils/user.ts

src/serializers/userProfile.ts
```diff
@@ -1,155 +0,0 @@
...
-              paymentStatus: true,
-            },
-          },
-        },
-      },
-    },
-  },
-} as const satisfies Prisma.UserSelect;
-
-export type UserProfilePayload = Prisma.UserGetPayload<{
```
timglorioso, line 105: Moved/renamed to UserSelectPayload in utils/user.ts

src/serializers/userProfile.ts
```diff
@@ -1,155 +0,0 @@
...
-  select: typeof USER_PROFILE_SELECT;
-}>;
-
-export function serializeUserProfile(
```
timglorioso, line 109: Merged behavior into user serializer, updated callers, added test

timglorioso:
@simonv3 thanks for your initial review. I made a few more changes to disambiguate the term "profile" until we have official listener profiles. The rest is mostly the same with minor touchups. See my comments above for the main changes. Next up will be several PRs with 1:1 renames which I hope will be much more boring and digestible.

src/routers/auth/index.ts
```diff
@@ -83,7 +86,21 @@ router.get("/logout", (req, res) => {
   }
 });
 
-router.get("/profile", userAuthenticated, profile);
+router.get("/profile", userAuthenticated, async (req, res, next) => {
```
simonv3: I personally prefer moving these into specific files. What's the reason for moving it back?

timglorioso: Mostly to disambiguate "profile". We could rename the file/handler instead.

timglorioso: I'm just going to revert and we can refine later.

simonv3: I didn't see the general comment sorry. Yeah I hear you, and I guess it's tricky to have the file handler not match the route name.

simonv3 (approved):

PR #2379, opened 2026-08-22
Fix album link opening in new tab in persistent player

### LINKED ISSUE #2375: Inconsistency between player links
**Describe the bug**
The bottom page player on mirlo.space has an inconsistency in how link related to the currently playing track are treated. Clicking on the artist name or the track name send you to the artist page and track page respectively, as expected, but the clicking on the album name opens up a new tab. This is inconsistent and i'm not sure if there is a reason for it.

**To Reproduce**
1. Play any track on mirlo.space
2. In the player, click on that track's album name

**Expected behavior**
The link should either open in the current tab, or show a relevant icon to signify it's going to open in a new tab.

**Desktop (please complete the following information):**
 - OS: Linux Mint 22.3 - Xfce 64-bit
 - Browser: Waterfow
 - Version 6.6.17 (64 bits)

theshreydhiman:
The artist and track title links in the bottom player bar navigate in the current tab, but the album name link opened a new tab because it was rendered through WidgetLink, which always sets target="_blank". That's correct when TrackTitleContent is used inside an embeddable widget, but the player passes useTrackArtistLinks to indicate it's part of the main Mirlo app, not a widget embed, and the other two links already respect that.

Route the album link through the same in-app Link component in that case, so all three links behave consistently. Widget embed contexts are unaffected since useTrackArtistLinks is never set there.

Fixes #2375

### DIFF
--- a/client/src/components/Widget/TrackTitleContent.tsx
+++ b/client/src/components/Widget/TrackTitleContent.tsx
@@ -35,7 +35,16 @@ export const TrackTitleContent: React.FC<{
     ? "Drafts"
     : track.trackGroup.title || tTrackGroup("untitled");
 
-  const trackGroupLink = (
+  // WidgetLink always opens in a new tab, which is correct when this content is
+  // rendered inside an embeddable widget (or a widget preview). The persistent
+  // player, however, sets useTrackArtistLinks to indicate we're navigating inside
+  // the main Mirlo app itself, where the artist and track title links already stay
+  // in the current tab — so the album link should match instead of standing out.
+  const trackGroupLink = useTrackArtistLinks ? (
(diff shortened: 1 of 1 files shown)


timglorioso:
Thanks for the contribution. I suggest changing the persistent player so it doesn't use widget components. It feels backwards to modify a widget component so that it works outside the context of a widget. What do you think?

Edit: on second thought, this seems like a harmless fix since the widget components are already used by the player.

theshreydhiman:
Thanks for taking a look. I agree the player shouldn't really be leaning on widget components — but untangling that felt like a bigger change than this issue needed, so I kept it to the one-line branch. Happy to go either way: leave this as-is, or close it and open something that moves the persistent player off WidgetLink properly. Which would you prefer?

timglorioso (approved):
This works, thanks again!

Issue #2377, opened 2026-08-22
Artist pictures won't update in label roster.


smonff:
An updated artist profile pictures won't update in it's label roster list. It updates fine on the artist page, though.

**To reproduce**

- Through a label account, upload an artist image
- Try and upload a different image
- Go to the label roster list

**Expected behavior**

The image should update immediately and there should not be a mismatch between the label roster and artist page if the artist is managed by a label.

**Example**

An example of this bug can be seen here:

- [Artist page with updated picture](https://mirlo.space/seb-hu-rillettes)
- [Label roster with old outdated pictures](https://mirlo.space/shr)

![Artist](https://github.com/user-attachments/assets/4c85a57a-80e7-4ce7-88c7-55f90b585edb)

![Roster](https://github.com/user-attachments/assets/d7897fa6-93de-41a0-9cf4-21d1624d2625)

**Environment**
 - OS: Ubuntu 24.04
 - Browser Firefox, Chromium (tried in both, also to make sure it was not a cache issue and I force-refreshed the cache)
- Tested on https://mirlo.space instance only, might be a server configuration thing

I waited a couple of days before reporting to make sure it was not a cache problem.

smonff:
It looks similar to #2362, but slightly different as the bug only occurs on the roster images.

smonff:
The left image is now correct, possibly a cache expiration effect.

<img width="1096" height="518" alt="Image" src="https://github.com/user-attachments/assets/22120b8a-2841-4412-99f9-89c9f434cc31" />

simonv3:
I think this should now be fixed.

Issue #2362, opened 2026-08-07
Changing album cover


Eulbast:
When I create a release, and that I publish it, if I want to upload a modified or new picture as cover, the old one stays printed. Inside the manage album page it is changed and updated, but outside it remains unchanged.
The only way to change it is to delete the whole album and redo the process from A to Z
Same thing for the profile logos, they don't refresh on my label page, for now I have to really be careful when I create a page to chose the perfect logo, no right to error when you already have several albums online. 
It would be great to gain flexibility on this side !

Thanks a lot

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser mozilla

Eulbast:
What I'd expect is to be able to test different photos according to my design, change it fluently, as it is now I can't take advantage of your tools- which are excellent- without those limitations. Changing the background photo works curiously like a charm, only covers and profile photos/ logo are a problem.

jayrope:
+1 - having the same problem here.

simonv3:
I think this is now fixed.

PR #2381, opened 2026-08-24
fix(images): version image URLs so re-uploads aren't served from cache

theshreydhiman:
Fixes the stale-image problem reported in #2362, and I think #2377 as well.

## What's happening

Re-uploading a cover, artist avatar or label logo writes to the same storage key, so `convertURLArrayToSizes` produces exactly the same URLs as before. Browsers and the CDN have no reason to refetch, so the old image keeps showing. The manage pages look right because they render the freshly uploaded file directly.

That also explains the detail in #2362 that background images *do* refresh: `ArtistBackground`, `UserBanner`, `ArtistHeaderSection` and `ArtistMerchListItem` already append `?${updatedAt}` by hand. Nothing else does.

## What this changes

`addSizesToImage` is the single place every serializer goes through for covers, avatars, banners, merch and subscription-tier images, and it already receives the image record. It now stamps that record's `updatedAt` onto each generated size URL, so every upload produces a distinct URL. No storage changes and no cache invalidation needed.

## Notes

The four components listed above will now append their own `?${updatedAt}` on top of the versioned URL. That is harmless, since the query string is opaque to the CDN and the URL still resolves, but those hand-rolled appends are redundant now and could come out in a follow-up. I left them alone to keep this to one function.

I wasn't able to run the stack locally, so this is reasoned from the code rather than verified against a running instance.

Happy to move the versioning into `convertURLArrayToSizes` instead, or scope it to covers only, if you'd prefer either.

### DIFF
--- a/src/utils/artist.ts
+++ b/src/utils/artist.ts
@@ -729,14 +729,38 @@ export const singleInclude = (queryOptions?: {
 
 export const addSizesToImage = (
   bucket: string,
-  image?: { url: string[] } | null
+  image?: { url: string[]; updatedAt?: Date | string } | null
 ) => {
-  return image
-    ? {
-        ...image,
-        sizes: image && convertURLArrayToSizes(image?.url, bucket),
-      }
-    : null;
+  if (!image) {
+    return null;
+  }
+
+  const sizes = convertURLArrayToSizes(image.url, bucket) as Record<
+    string,
+    string
+  >;
+
+  // Re-uploading an image reuses the same object key, so its URLs never change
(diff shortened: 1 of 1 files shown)


simonv3:
Thanks @theshreydhiman !! Wanna open a ticket or DM me on discord to see about getting the stack running locally?

simonv3 (approved):

simonv3:
Merged! I think in the future it'd also be great to add some API tests that check that this updatedAt gets appended to the image sizes.

PR #2382, opened 2026-08-25
Artist-to-Profile Internal 10: purchase locals

timglorioso:
Endpoints:

/artists/{id}/purchaseCatalogue
/artists/{id}/subscribe
/checkout
/manage/merch/{merchId}
/manage/purchases/{purchaseId}/contactArtist
/manage/purchases/{purchaseId}
/merch/{id}
/purchase
/purchase/{id}
/stripe/setupIntentStatus


src/utils/payments/purchase.ts
```diff
@@ -110,7 +110,7 @@ export const initiatePayment = async ({
   const metadata: Record<string, string> = {
     purchaseType,
     stripeAccountId,
-    artistId: String(artistId),
+    artistId: String(profileId),
```
simonv3, line 113: Could we add both for redundancy? `profileId` and `artistId`?

simonv3 (dismissed):
Looks good to me! Just one thought about sending stuff to Stripe and being a little safer but it looks fine?

src/utils/payments/purchase.ts
timglorioso: the actual metadata key is still artistId

src/utils/payments/PaymentProcessor.ts
timglorioso: missed lines 27 and 126

src/utils/payments/purchase.ts
simonv3: Gotcha!  I'm wondering if, for example, for subscriptions or re-attempts in the future, whether it'll keep the original artistId? Like, if a subscription's metadata has it, and then we change it, and a month later it's looking for artistId still. I'm not sure if that matters because I'm not sure we actually use the metadata on a re-run?

simonv3 (approved):

src/utils/payments/purchase.ts
timglorioso: As far as I can tell, we have a couple options:
1. Migrate the metadata and all the code to use profileId (phased, first code reads/writes both, migrate metadata to profileId, then switchover code)
2. Don't migrate metadata, start writing profileId, and always read for both

timglorioso: @simonv3 my sense here is what we do with metadata can be considered later, and these renames are okay as-is. Am I missing something?

PR #2396, opened 2026-09-11
Serve app link files for the mobile app

peterklingelhofer:
Related: https://github.com/funmusicplace/mirlo-mobile/pull/75

### DIFF
--- a/src/wellKnown.ts
+++ b/src/wellKnown.ts
@@ -8,12 +8,99 @@ router.get("/.well-known/webfinger", (req, res) => {
   res.status(404).json({ error: "Not found" });
 });
 
-router.get("/.well-known/assetlinks", (req, res) => {
+// Android App Links and iOS Universal Links for the mobile app (space.mirlo.mobile).
+// Both stores' keys are listed: Play re-signs store installs with its own key,
+// EAS signs everything else with the upload key.
+const assetLinks = [
+  {
+    relation: ["delegate_permission/common.handle_all_urls"],
+    target: {
+      namespace: "android_app",
+      package_name: "space.mirlo.mobile",
+      sha256_cert_fingerprints: [
+        "C1:77:8F:94:E9:A5:25:8A:6C:67:6D:38:0E:31:BA:F7:49:41:D5:9A:13:45:75:94:2F:16:01:B9:C3:E6:8E:BB",
+        "17:56:49:BC:2D:8F:32:26:B8:B8:A6:2E:29:81:44:18:7B:A1:BE:9A:4B:30:54:6E:C4:B7:41:42:9F:1E:1E:1A",
+      ],
+    },
+  },
+];
+
+// Top-level routes in client/src/routes.tsx that sit beside artist slugs, and
+// artist sub-pages the app has no screen for. Everything else under / opens the app.
+const webOnlyRoots = [
+  "account",
+  "admin",
+  "artists",
+  "checkout",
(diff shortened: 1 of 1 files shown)


src/wellKnown.ts
```diff
@@ -8,12 +8,99 @@ router.get("/.well-known/webfinger", (req, res) => {
...
+        "C1:77:8F:94:E9:A5:25:8A:6C:67:6D:38:0E:31:BA:F7:49:41:D5:9A:13:45:75:94:2F:16:01:B9:C3:E6:8E:BB",
+        "17:56:49:BC:2D:8F:32:26:B8:B8:A6:2E:29:81:44:18:7B:A1:BE:9A:4B:30:54:6E:C4:B7:41:42:9F:1E:1E:1A",
+      ],
+    },
+  },
+];
+
+// Top-level routes in client/src/routes.tsx that sit beside artist slugs, and
+// artist sub-pages the app has no screen for. Everything else under / opens the app.
+const webOnlyRoots = [
```
simonv3, line 30: dang is there no way this can be future safe? the issue is if we add a new route we'd have to add it here, but I guess because we start with "/:artistSlug" which is all consuming it's tricky?

src/wellKnown.ts
```diff
@@ -8,12 +8,99 @@ router.get("/.well-known/webfinger", (req, res) => {
...
+  "connect",
+  "links",
+  "merch",
+  "merch/*",
+  "posts",
+  "posts/*",
+  "release/*/download",
+  "release/*/redeem",
+  "release/*/tracks/*/download",
+  "releases",
```
simonv3, line 65: Wouldn't this be the same as the artist album page and so show the artist page in the app? dunno

src/wellKnown.ts
peterklingelhofer: Inverted it. The exclusion list of web routes is gone from the app: a single segment is treated as an artist only if GET /v1/artists/:slug succeeds (the fetch also primes the artist page), and a 404 opens the page in an in-app browser tab. New routes need nothing on that side

On the AASA I kept a short list, but by a criterion rather than the router: paths an in-app browser sheet would break, because on iOS the sheet has its own cookie jar and no download UI. That's auth, account, admin, checkout, and the download/redeem/unsubscribe links from emails. Content routes aren't listed on purpose. So a new content route needs nothing here either; a new auth route is one line.

/<artist>/releases now opens the artist page in the app on both platforms, good call.

The underlying wrinkle is artists living at /. If the API doesn't already reject route names as artist slugs, that's the durable fix. What do you think? Either way it's clunky but this may be the best / most maintainable option of some clunky options

src/wellKnown.ts
peterklingelhofer: Addressed this in the latest commits:

**Web**: releases is no longer excluded in the AASA, so iOS's /* sends it to the app

**Android**: added the /[^/]+/releases pattern to [app.json](vscode-webview://0amnu73jtqdaj8r4kovdeckn5pbpk149bv4tihq7tccd2t0ih1j1/app.json), since Android only captures what the manifest lists

**App**: [+native-intent.tsx](vscode-webview://0amnu73jtqdaj8r4kovdeckn5pbpk149bv4tihq7tccd2t0ih1j1/app/+native-intent.tsx) treats releases like the bare artist path (!section || section === "releases"), after the API check

simonv3 (approved):

PR #2402, opened 2026-09-16
Flagged content feed

LouisLouiseKay:
This adds a flagged content feed, that lives outside the general instance admin panel (and is readily accessible in the sidebar menu if you have admin rights), with the idea that this could be a page opened to trusted users later down the road in more community oriented instances. This also fixes a bug which has been picked upon recently where a page would remain visible if you had access to its URL, despite it being disabled by the content moderation team.


src/routers/v1/admin/contentFlags/{id}.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      const updated = await prisma.contentFlag.update({
+        where: { id: flag.id },
+        data: resolved
+          ? {
+              resolvedAt: new Date(),
+              resolvedByUserId: req.user.id,
+              reporterEmail: null,
+            }
+          : { resolvedAt: null, resolvedByUserId: null },
```
simonv3, line 48: if the body doesn't contain `resolved` it would unresolve?

src/routers/v1/flag.ts
```diff
@@ -44,6 +44,17 @@ export default function () {
         }
       }
 
+      await prisma.contentFlag.create({
```
simonv3, line 43: someone who enters a random string to this endpoint could get to this point and we'd be creating data reports for trackgroups that don't exist?

src/utils/contentFlag.ts
```diff
@@ -0,0 +1,71 @@
...
+      return { profileId: background?.profileId };
+    }
+    case "postImage": {
+      const image = await prisma.postImage.findUnique({
+        where: { id: imageId },
+        select: { post: { select: { profileId: true } } },
+      });
+      return { profileId: image?.post.profileId ?? undefined };
+    }
+    case "merchImage": {
```
simonv3, line 61: will we care if `userAvatar` `userBanner` or `image` are ever AI generated or offensive in some other form? Realizing that the hope (dream, aspiration) is that we build a generic `image` class eventually.

src/routers/v1/trackGroups/{id}/index.ts
```diff
@@ -32,6 +33,7 @@ export default function () {
         trackGroup = await prisma.trackGroup.findFirst({
           where: {
             id: actualId,
+            ...(loggedInUser?.isAdmin ? {} : whereForVisibleTrackGroup()),
```
simonv3, line 36: should this same kind of where be included for merch?

src/routers/v1/tracks/{id}/stream/{segment}.ts
```diff
@@ -34,8 +35,11 @@ export default function () {
     const user = req.user;
 
     try {
-      const track = await prisma.track.findUnique({
-        where: { id: Number(id) },
+      const track = await prisma.track.findFirst({
+        where: {
+          id: Number(id),
+          ...(user?.isAdmin ? {} : { trackGroup: whereForVisibleTrackGroup() }),
```
simonv3, line 41: Should probably also do this in stream/external (the fairplayer / canimus route)

src/routers/v1/flag.ts
LouisLouiseKay: Hadn't thought of that 😬

src/routers/v1/trackGroups/{id}/index.ts
LouisLouiseKay: Yes, missed that.

src/routers/v1/tracks/{id}/stream/{segment}.ts
LouisLouiseKay: Ah yes, that's actually quite important, thanks for reminding me.

src/utils/contentFlag.ts
LouisLouiseKay: I feel like we will, but I wasn't too sure how actionable this was for now.

simonv3 (approved):

